PR #33, opened 2016-02-11
Store snapshot objects in C++ plot object

huguesv:
DON'T MERGE UNTIL BUG IS APPROVED: https://github.com/Microsoft/RTVS/issues/886

Stop unnecessarily saving the snapshot objects in the R session.  It's sufficient to keep the SEXP in the C++ plot object.

Fix implementation of equal operator of protected_sexp, and remove the one that was unused.  The way they were implemented previously, they kept calling each other until you got an exception.


huguesv:
I will resubmit this from my newly recreated fork.

PR #36, opened 2016-02-17
Create a trimmed log file

int19h:


int19h:
Fixed by other means.

PR #37, opened 2016-02-18
Fix crashing on evaluation requests that contain code with syntax errors

int19h:


MikhailArkhipov:
VS fix is probably safer.

MikhailArkhipov:
But otherwise :+1

PR #38, opened 2016-02-18
Snapshot storage

huguesv:
Store snapshots in c++ objects, and fix issue in equal operator of protected_sexp.
This is an identical resubmission of https://github.com/Microsoft/R-Host/pull/33 from my new fork.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -103,11 +103,8 @@ namespace rhost {
             using unique_ptr::operator=;
 
             protected_sexp& operator= (SEXP other) {
-                return *this = std::move(protected_sexp(other));
-            }
-
-            protected_sexp& operator= (const protected_sexp& other) {
-                return *this = other.get();
+                swap(protected_sexp(other));
+                return *this;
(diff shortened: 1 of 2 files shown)


src/util.h
```diff
@@ -103,11 +103,8 @@ namespace rhost {
             using unique_ptr::operator=;
 
             protected_sexp& operator= (SEXP other) {
```
int19h: This class should still have `protected_sexp& operator= (const protected_sexp& other)`, so please don't remove that. The `swap` change should take care of infinite recursion otherwise.

Also, it would be good to add ` operator= (protected_sexp&& other)` - it can just do `swap` as well. This will make containers of `protected_sexp` more efficient, because e.g. `vector` wouldn't add/remove references to every single element every time it has to resize.

huguesv:
Thanks!  Updated to add back the deleted assignment operator and added the assignment move operator.

int19h:
:+1:

PR #39, opened 2016-02-18
Relax graphics engine version check to allow R 3.2 and 3.3

huguesv:
I've verified that the public graphics engine/device structs that we compile against haven't changed in R 3.3, so we can run on both 3.2 and 3.3 without any conditional code.

Edit: Removed don't merge warning, since this is for public preview milestone.


int19h:
:+1:

PR #41, opened 2016-02-21
Fix fancy quote conversion with dQuote/sQuote

MikhailArkhipov:


src/util.cpp
```diff
@@ -56,6 +56,40 @@ namespace rhost {
...
+                            break;
+                        case 0x0093:
+                            wc[j] = 0x201C; // Left double quote
+                            break;
+                        case 0x0094:
+                            wc[j] = 0x201D; // Right double quote
+                            break;
+                    }
+                }
+                i += char_size - 1;
```
AlexanderSher: A bit hard to read. I think it would be simpler if you remove `i++` from loop definition and here it would be `i += char_size`

int19h:
I think this is the wrong place to do it, because it is a lossy data conversion, and will subtly break various scenarios such as editing variable values (basically the moment you edit it, even if you do not change it, the quotes will change). It should be done on VS side, and as a very last step before presenting it in UI (i.e. not in `RHost` or `RSession` either, but when we print it out to the REPL or display it in WPF).

MikhailArkhipov:
I can be done on VS side but then every place that may be using string result for representation will have to be aware of this case. One point of conversion may be more stable. Conversion is not truly lossy it is more like 'correcting' :-) since no one really should be using quotes in 0x90 range...

int19h:
The other option is to change `options("useFancyQuotes")` to `FALSE`, so that it does not produce those fancy quotes to begin with.

int19h:
My main concern is that it's not a given that every single string that goes from host to VS will actually be used for representation purposes. So, I'd rather have as close to 100% fidelity on the wire, so that any particular component can decide what to do with the data once it has it.

MikhailArkhipov:
They can always change it back. I'll see how many place over in VS may have to have the conversion.

MikhailArkhipov:
It is all over: console, VE, grid viewer (ros/col names, values), locals, watch

int19h:
Locals shouldn't do it precisely because it means that every edit operation will be silently changing things without user's knowledge, even when they think they didn't change anything, or changed something unrelated. Invisible quotes are vastly preferable to that.

Generally speaking, we should not be changing state silently like that - it is a rather fundamental requirement for a debugger. Think what happens if a string is used to store binary data, for example, and we suddenly swap some bytes in it. 

Changing back is not reliable - by the time you get to it, you don't know if substitution already happened or not - i.e. you see a Unicode double quote character, for example, and you don't know if it was like that, or it was substituted. Reliable round-tripping requires full preservation of original data.

MikhailArkhipov:
But if we show one value in VE and another in Locals users will get confused - we will get bugs in this. What is they want to remove quotes in Locals or Watch?

int19h:
There are other invisible characters that are not editable in a similar manner, like non-breaking space or soft hyphen.

The ideal fix would be to do the conversion right before displaying/editing the value, but remember which characters were converted, and then converting them back after the edit if they were touched. But to implement this later, the data has to be untouched anywhere in the pipeline all the way until it gets to the view layer - hence my reluctance to take this fix. It is cheap here and now, but it prevents everything else down the line from doing this right later unless it's taken out completely (which means that everything else down the line has to be fixed at once... which probably means that we'll never do it because it's too expensive).

Yes, doing it in VS means that we'll need to convert (two-way for Locals, one-way everywhere else for now) in all those places. But it's not really hard with a simple helper extension method. And it can then be changed to do the right thing on a case-by-case basis.

int19h:
Feel free to create & assign the the quote issue with Locals specifically (where it needs to be two-way) to me.

MikhailArkhipov:
That cases by case worries me. There may be bug in every place that we may need to convert. RS displays everything and will allow editing where appropriate. Why convert in 6 places when we can have it in one?

MikhailArkhipov:
I actually wrote the said extension and planted it already in 3 places and that's not all... Grid needs it as does Excel conversion and when we get 2 way with Excel, that will be another and when we get to database connections there will be more.

int19h:
RS does not allow editing in their Variable Explorer equivalent.

If you want to do it in one place, let's do it in `DebugValueEvaluationResult.Representation`. We can have `Representation` provide the converted values, and add `RawRepresentation` or something like that to give access to raw data.

But it definitely should not be in the host, because that blocks VS from getting the raw data at all, and that is a deal breaker.

MikhailArkhipov:
OK. I didn't want to change DebugValueEvaluationResult.Representation but we can do it if that's OK. I'll add Raw then. There is still going to be an issue if anyone gets strings by raw eval though.

int19h:
Hence the suggestion to make it a method with an explicit argument, e.g.:
```cs
enum DebugValueRepresentationKind {
   Normal = 0,
   Raw = 1,
   ...
}

string GetRepresentation(DebugValueRepresentationKind kind) { ... }

string s = valueResult.GetRepresentation(DebugValueRepresentationKind.Raw);
```

That way you _must_ specify what you want every time, and so it forces the caller to have to think about which is the best for any particular case. And the default would be to handle quotes.

The more I think about this, the more I like it, because we can also put Unicode escaping/unescaping there in a similar way, and in the future, any other transformations.

Later on, I'll also add flags for `Deparse`, `Str` and `ToString`, and remove `DebugValueEvaluationResultRepresentation` struct completely.

int19h:
Another thing, I can later optimize it and make `GetRepresentation` async and have it fetch the value if it doesn't have it yet. Right now there is already one place where we fetch one representation first, and then later fetch a different one - debugger normally uses `Deparse` everywhere, but it needs `ToString` for text/HTML/XML/JSON viewer. Currently what it does is re-evaluate the same expression and get a new eval result with a new representation; but if `GetRepresentation` can encapsulate that, it's even better.

int19h:
For now though let's just do the min fix with a non-async `GetRepresentation` that returns the existing struct - it could even just take a `bool` to specify if it's raw or not instead of enum. I'll add all this other stuff later when we have more time to polish things.

MikhailArkhipov:
OR

        public DebugValueEvaluationResultRepresentation(JObject repr) {
            Deparse = repr.Value<string>("deparse");
            ToString = repr.Value<string>("toString");
            RawStr = repr.Value<string>("str");
            Str = RawStr.ToUnicodeQuotes();
        }

int19h:
It feels a bit weird to have `Str`/`RawStr` but not others similarly paired. Also, Locals uses `Deparse` almost everywhere.

MikhailArkhipov:
Task IRCallbacks.WriteConsoleEx(string buf, OutputType otype, CancellationToken ct) {
            Output?.Invoke(this, new ROutputEventArgs(otype, buf.ToUnicodeQuotes()));

?

MikhailArkhipov:
Or at the receiving end of the event?

int19h:
On the receiving end would be preferable (so that subscribers can decide).

PR #42, opened 2016-02-22
Fix #739: Need to sign Microsoft.R.Host.exe in builds

AlexanderSher:
RHost part of fix

### DIFF
--- a/src/Microsoft.R.Host.vcxproj
+++ b/src/Microsoft.R.Host.vcxproj
@@ -204,6 +204,11 @@
   <ItemGroup>
     <ResourceCompile Include="Resource.rc" />
   </ItemGroup>
+  <ItemGroup>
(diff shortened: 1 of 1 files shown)


src/Microsoft.R.Host.vcxproj
```diff
@@ -204,6 +204,11 @@
   <ItemGroup>
     <ResourceCompile Include="Resource.rc" />
   </ItemGroup>
+  <ItemGroup>
+    <FilesToSign Include="$(OutDir)\RHost.exe ">
```
MikhailArkhipov: Is this a real name?

int19h: Should be Microsoft.R.Host.exe

int19h:
But otherwise :+1:

PR #43, opened 2016-02-22
Fix #917: memory.limit isn't set in some cases

int19h:
Use GlobalMemoryStatusEx (like Rterm) rather than GetPhysicallyInstalledSystemMemory to retrieve the amount of physical memory on the system - the latter function fails on VMs.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -116,14 +116,21 @@ namespace rhost {
 
     void set_memory_limit() {
 #ifdef WIN32
-        ULONGLONG total;
-        if (!GetPhysicallyInstalledSystemMemory(&total)) {
+        MEMORYSTATUSEX ms = {};
+        ms.dwLength = sizeof ms;
+        if (!GlobalMemoryStatusEx(&ms)) {
+            logf("Couldn't set R memory limit - GlobalMemoryStatusEx failed with GetLastError=%u\n", GetLastError());
             return;
         }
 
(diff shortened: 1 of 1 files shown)


MikhailArkhipov:
:+1:

AlexanderSher:
:+1:

src/main.cpp
```diff
@@ -116,14 +116,21 @@ namespace rhost {
...
-        ULONGLONG total;
-        if (!GetPhysicallyInstalledSystemMemory(&total)) {
+        MEMORYSTATUSEX ms = {};
+        ms.dwLength = sizeof ms;
+        if (!GlobalMemoryStatusEx(&ms)) {
+            logf("Couldn't set R memory limit - GlobalMemoryStatusEx failed with GetLastError=%u\n", GetLastError());
             return;
         }
 
+        double memory_limit = static_cast<double>(ms.ullTotalPhys / 1024 / 1024);
```
huguesv: You won't get a fractional part with this.  You need to have one of the 1024 be a double.

huguesv: Maybe that's what you intended, in that case never mind :)

int19h: Yep, but should never actually be the case - I've yet to see a machine with physical memory not measured in whole megabytes :)

And even should it happen, rounding down is safe here. It's only a `double` because that's what R expects for this function (presumably to allow for bigger range, and because R doesn't have a 64-bit integer type). Internally, `in_memsize` will convert it back to bytes, and store it in a 64-bit int.

PR #44, opened 2016-02-23
Add clear all plots and remove active plot functionality.

huguesv:


src/grdeviceside.cpp
```diff
@@ -356,7 +360,19 @@ namespace rhost {
...
+                _active_plot = _plots.begin();
+            }
+
+            void plot_history::remove_active() {
+                if (_active_plot != _plots.end()) {
+                    _active_plot = _plots.erase(_active_plot);
+                    // erase() returns an iterator that points to end() when removing the last item
+                    // so adjust it to point to the new last item, if one is available
+                    if (_active_plot == _plots.end() && _plots.size() > 0) {
+                        _active_plot = std::prev(_plots.end(), 1);
```
int19h: You can just do `--_active_plot` here instead of `std::prev`, since it is a vector, and vector iterators are bidirectional and random-access.

int19h:
:+1:

PR #45, opened 2016-02-23
Fix #912: Step Out at top level doesn't work [R-Host]

int19h:
Add support for skipping toplevel contexts to browser_set_debug, so that it can be used in an evaluation request (which always wraps the eval in a toplevel context).


huguesv:
:+1:

PR #46, opened 2016-02-25
Fix #958: CRAN R startup message does not appear in RTVS

int19h:
Disable quiet mode.

Also add startup banner for MRO per #973.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -27,11 +27,13 @@
 #include "host.h"
 #include "Rapi.h"
 #include "util.h"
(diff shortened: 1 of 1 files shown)


huguesv:
There's a space after the end of the sentence, before the new lines, not sure if it's intended or not.

:+1:

MikhailArkhipov:
:+1:

int19h:
The space is because C++ editor makes `\n` part of the URL otherwise. Doesn't affect anything in the REPL so I kept it.

PR #47, opened 2016-02-26
Host crash on navigation to invalid plot

huguesv:
Check for nullptr on the plot snapshot. Remove warning about plot history corruption, as that was for the case where the user tempered with the internal session variables which don't exist anymore.

Fix for https://github.com/Microsoft/RTVS/issues/1021


int19h:
:+1:

PR #49, opened 2016-03-22
Make evaluation requests asynchronous

int19h:
Make evaluation requests asynchronous (i.e. they are permitted at any point, just like cancellation requests), and handle them during `CallBack`, in addition to other R callbacks.

This is the host-side infrastructure necessary to fix these:

https://github.com/Microsoft/RTVS/issues/541
https://github.com/Microsoft/RTVS/issues/547

However, it does not change the semantics of any existing eval calls, and so is safe to commit without any changes to VS side.

### DIFF
--- a/src/host.cpp
+++ b/src/host.cpp
@@ -32,795 +32,834 @@ using namespace rhost::util;
 using namespace rhost::eval;
 
 namespace rhost {
-    namespace host {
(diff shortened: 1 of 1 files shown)


int19h:
Because some functions were moved around, Git diff is broken. I'll start a CodeFlow review.

PR #50, opened 2016-03-29
Fast JSON serialization

int19h:
(This is for https://github.com/Microsoft/RTVS/issues/1350)

Add a fast builtin R-to-JSON serializer, and change the semantics of =j eval to implicitly use that serializer on the returned value (instead of accepting a serialized string and parsing it).

Expose serializer to R as a .Call function.

Add missing MicroBuild NuGet package reference.

### DIFF
--- a/src/json.cpp
+++ b/src/json.cpp
@@ -166,7 +166,14 @@ namespace rhost {
             case REALSXP: {
                 at_most_one(sexp);
                 double x = *REAL(sexp);
-                result = R_IsNA(x) ? js::value() : js::value(x);
+                if (R_IsNA(x)) {
+                    result = js::value();
+                } else {
+                    if (isinf(x) || isnan(x)) {
+                        json_error(sexp, "+Inf, -Inf and NaN cannot be serialized");
+                    }
+                    result = js::value(x);
+                }
                 break;
             }
 
--- a/src/json.cpp
+++ b/src/json.cpp
@@ -118,12 +118,9 @@ namespace rhost {
         }
 
         bool to_json(SEXP sexp, js::value& result) {
-            if (Rf_length(sexp) == 0) {
-                result = js::value();
-                return true;
-            }
+            int type = TYPEOF(sexp);
 
-            switch (TYPEOF(sexp)) {
+            switch (type) {
             case NILSXP:
                 result = js::value();
                 return true;
@@ -144,7 +141,14 @@ namespace rhost {
                 result = js::value(js::object());
                 env_to_object(sexp, result.get<js::object>());
                 return true;
+            }
 
+            if (Rf_length(sexp) == 0) {
+                result = js::value();
+                return true;
+            }
+
+            switch (type) {
             case LGLSXP: {
                 at_most_one(sexp);
                 int x = *LOGICAL(sexp);
--- a/src/json.cpp
+++ b/src/json.cpp
@@ -1,0 +1,190 @@
+/* ****************************************************************************
+ *
+ * Copyright (c) Microsoft Corporation. All rights reserved.
+ *
+ *
+ * This file is part of Microsoft R Host.
+ *
+ * Microsoft R Host is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * Microsoft R Host is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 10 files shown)


src/json.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+            case ENVSXP:
+                result = js::value(js::object());
+                env_to_object(sexp, result.get<js::object>());
+                return true;
+
+            case LGLSXP: {
+                at_most_one(sexp);
+                int x = *LOGICAL(sexp);
+                result = x == R_NaInt ? js::value() : js::value(x != 0);
```
MikhailArkhipov: I think there is a ```#define NA_LOGICAL``` somewhere, but it is ```R_NaInt``` anyway

src/json.cpp
```diff
@@ -0,0 +1,190 @@
...
+                at_most_one(sexp);
+                int x = *INTEGER(sexp);
+                result = x == R_NaInt ? js::value() : js::value(static_cast<double>(x));
+                break;
+            }
+
+            case REALSXP: {
+                at_most_one(sexp);
+                double x = *REAL(sexp);
+                result = R_IsNA(x) ? js::value() : js::value(x);
```
MikhailArkhipov: R_NaReal ?

MikhailArkhipov:
Needs tests
:+1:

src/r_util.cpp
```diff
@@ -434,6 +432,15 @@ namespace rhost {
             return R_NilValue;
         }
 
+        extern "C" SEXP toJSON(SEXP obj) {
+            SEXP json = Rf_mkCharCE(to_json(obj).serialize().c_str(), CE_UTF8);
```
huguesv: Should you use rhost::util::exceptions_to_errors to protect against exceptions from `to_json`, as it uses picojson and std which may throw?

src/json.h
```diff
@@ -0,0 +1,52 @@
...
+        //
+        // NULL, NA, empty vector -> null
+        // TRUE -> true
+        // FALSE -> false
+        // Vector of a single non-NA integer or double -> numeric literal
+        // Vector of a single non-NA string -> string literal
+        // List with all elements unnamed -> array (recursively)
+        // List with all elements named, or environment -> object (recursively)
+        //
+        // If any element of a list or environment is NA, that element is skipped.
```
huguesv: Can you explain why we have to skip NA elements?

While that may not be a problem for data that we create ourselves (we can avoid putting NA elements in the first place), if we want to serialize a list that is created by the user or R library code, we'll get unexpected results.

src/json.cpp
int19h: I've looked at `R_IsNA`, and it does a little bit more than that. I think you're supposed to use `R_NaReal` if you want to create a value, but you're supposed to use the function for checks. At least that's what R code does everywhere.

src/json.cpp
int19h: In the standard R headers, yes. There isn't one in ours precisely because it's just an alias, so you're just supposed to use `R_NaInt` directly in this case.

src/json.h
int19h: It is a shortcut to simplify building up large data structures in R code. Basically, it lets you preallocate a list for N elements, but then make "holes" in it as you scan your data and decide that some elements are not applicable. It also allows the convenient `NA_if_error` function to be used without any additional checks - you can see that a lot in `eval.R`.

(Keep in mind that this is not meant to be a general serialization solution for arbitrary R data structures, hence various constraints that it places on input. If you are passing a data structure directly from some R library to this serializer, you need to know very well what exactly the shape of that data is, otherwise you'll likely run afoul of other rules like no multiple-element vectrors, anyway.)

Having said that, I do somewhat dislike the use of `NA` specifically as sentinel. It is true that it's much more likely to appear in valid data, and it can be convenient to have it represented. I would like to replace that sentinel with an empty vector - `c()` - instead. However, I wanted to keep the behavior exactly the same as the old R serializer for now, since existing code in debugger relies on it, and that includes `NA` handling. Changes to improve the mapping can come later.

src/r_util.cpp
int19h: Sure.

src/json.h
int19h: Another better option for a sentinel in the future might be a special singleton environment. That way it won't ever conflict with any value any existing R code may produce.

huguesv:
:+1:

int19h:
The fixes above came from tests. Tests are implemented on RTVS side, because we don't have an infrastructure for C++ tests currently. They're part of this review: https://github.com/Microsoft/RTVS/pull/1368

PR #51, opened 2016-03-30
Make JSON serialization errors non-fatal

int19h:
Make JSON serialization errors non-fatal in the serializer itself (failing to serialize JSON result for `=j` eval is still fatal, but errors from `toJSON` can now be handled).

Reorganize JSON serializer error messages for better readability.


huguesv:
:+1:

PR #52, opened 2016-04-08
Fix #1493: BeginInteractionAsync deadlock

int19h:
When receiving messages, do not overwrite the global variable used to store request response with eval and cancellation messages. Also, refactor code, renaming variables and adding comments to make it clearer which is used where, and separating mutexes for various unrelated things.

Disable `--rhost-listen` (host server mode) until that code path is properly analyzed and vetted on security.

Add VS2015 U2 C++ Intellisense database files to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,3 +83,5 @@ UpgradeLog*.htm
 *.opensdf
 src/GeneratedVersion.h
 src/Microsoft.R.Host.VC.opendb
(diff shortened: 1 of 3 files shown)


int19h:
`--rhost-listen` change is not related to the bug; it's just something that @jflam asked to do for this release, and this was a convenient change to bundle it with.

MikhailArkhipov:
:+1:

PR #54, opened 2016-04-11
Fix #1513: Variable names with non-ASCII characters produce errors

int19h:
(https://github.com/Microsoft/RTVS/issues/1513)

Convert names to UTF-8 when generating JSON.

### DIFF
--- a/src/json.cpp
+++ b/src/json.cpp
@@ -30,17 +30,17 @@ namespace js = picojson;
 namespace rhost {
     namespace json {
         namespace {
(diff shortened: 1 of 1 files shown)


src/json.cpp
```diff
@@ -30,6 +30,18 @@ namespace js = picojson;
 namespace rhost {
     namespace json {
         namespace {
+			bool strsxp_to_utf8(SEXP sexp, std::string& result) {
```
huguesv: something wrong with the indentation

huguesv:
:+1:

PR #55, opened 2016-04-12
Fix #1340: Non-ASCII output is garbled when R locale is not Latin-1

int19h:
Use mbstowcs and wctomb from MinGW, rather than from VC, when converting to/from UTF-8, so as to respect MinGW (and hence R) locale.


huguesv:
:+1:

PR #56, opened 2016-04-15
Fix #1562: Sometimes they never die...

int19h:
Perform heartbeat pings to client periodically once connection is established, and shut the host down if client does not respond.


int19h:
https://github.com/Microsoft/RTVS/issues/1562

huguesv:
:+1:

PR #58, opened 2016-04-15
Don't force create file device when resizing, to prevent creating an …

huguesv:
…empty png file that isn't tracked.

Fixes https://github.com/Microsoft/RTVS/issues/1568


int19h:
:+1:

PR #59, opened 2016-04-15
Add an evaluation request flag to specify that the value produced by the expression is not needed (and thus avoid having to serialize and transmit it).

int19h:


huguesv:
:+1:

PR #60, opened 2016-04-20
Fix https://github.com/Microsoft/RTVS/issues/1629

int19h:
Do not improperly reset response state when eval completes.

### DIFF
--- a/src/host.cpp
+++ b/src/host.cpp
@@ -358,11 +358,6 @@ namespace rhost {
                 }
             }
 
-            {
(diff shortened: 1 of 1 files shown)


int19h:
To clarify the fix - this code is leftover from when evals were handled in the same code path as other regular requests/responses. Since evals are fully async now, they should never use this flag at all - it's purely for R callbacks, and responses to those callbacks. That this code was left when async evals were checked in was my oversight.

huguesv:
👍

PR #62, opened 2016-04-27
Fix https://github.com/Microsoft/RTVS/issues/1701

int19h:
(Stop Debugging after Break crashes host process)

When entering wait-for-response state after sending a request, store the previous state, and restore it upon receiving the response, so that any outer request awaiting a response remains in proper state.


MikhailArkhipov:
:+1:

PR #63, opened 2016-04-27
Implement graphics device locator functionality (r-host)

huguesv:


src/grdeviceside.cpp
```diff
@@ -508,7 +508,19 @@ namespace rhost {
             Rboolean ide_device::locator(double *x, double *y) {
                 *x = 0;
                 *y = 0;
-                return R_FALSE;
+                Rboolean clicked = R_FALSE;
+
+                rhost::host::with_cancellation([&] {
+                    auto msg = rhost::host::send_message_and_get_response("Locator");
+                    auto& result_clicked = msg.args[0].get<bool>();
```
int19h: Can you add some validation code here with `is<...>` checks to make sure that message is not malformed, and `fatal_error` if it is?

huguesv:
Added response validation.

int19h:
:+1:

PR #65, opened 2016-05-26
Plot

huguesv:
This allows using ExecuteAsync for plot commands (next, previous, delete, resize) instead of interactions, by making sure that the plot is sent to the IDE before the R functions for those commands return.

Changes the plot message to include the state (active, count), to avoid having the IDE call back immediately to query that info.

Changes are required on the VS side, I will wait to merge until those are ready.

### DIFF
--- a/src/grdeviceside.cpp
+++ b/src/grdeviceside.cpp
@@ -931,27 +931,6 @@ namespace rhost {
                 });
             }
 
-            extern "C" SEXP ide_graphicsdevice_history_info(SEXP args) {
-                return rhost::util::exceptions_to_errors([&] {
-                    // zero-based index of active plot, number of plots
-                    auto plotIndex = Rf_allocVector(INTSXP, 1);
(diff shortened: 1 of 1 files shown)


src/grdeviceside.cpp
```diff
@@ -708,7 +708,12 @@ namespace rhost {
             void ide_device::send(const std::tr2::sys::path& filename) {
                 auto path_copy(filename);
                 rhost::host::with_cancellation([&] {
-                    rhost::host::send_message("Plot", rhost::util::to_utf8(path_copy.make_preferred().string()));
+                    rhost::host::send_message(
+                        "Plot",
+                        rhost::util::to_utf8(path_copy.make_preferred().string()),
+                        (double)active_plot_index(),
```
int19h: Can you change these to `static_cast` for clarify?

int19h:
:+1:

PR #67, opened 2016-06-16
Fix #1897: Remove CRT dependency for Dev15

int19h:
Statically link the runtime.

### DIFF
--- a/src/Microsoft.R.Host.vcxproj
+++ b/src/Microsoft.R.Host.vcxproj
@@ -101,6 +101,7 @@
       <PreprocessorDefinitions>WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <SDLCheck>false</SDLCheck>
       <AdditionalIncludeDirectories>..\lib\picojson;..\lib\websocketpp</AdditionalIncludeDirectories>
+      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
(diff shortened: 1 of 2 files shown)


src/host.cpp
```diff
@@ -561,7 +561,7 @@ namespace rhost {
             }
         }
 
-        extern "C" int ReadConsole(const char* prompt, char* buf, int len, int addToHistory) {
+        extern "C" int R_ReadConsole(const char* prompt, char* buf, int len, int addToHistory) {
```
int19h: This is needed because it clashes with Win32 `ReadConsole` from kernel32.lib during linking. I'm not sure why static vs dynamic linking of CRT makes a difference here, but either way it's easy to fix by renaming our symbol.

src/Microsoft.R.Host.vcxproj
```diff
@@ -101,6 +101,7 @@
       <PreprocessorDefinitions>WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <SDLCheck>false</SDLCheck>
       <AdditionalIncludeDirectories>..\lib\picojson;..\lib\websocketpp</AdditionalIncludeDirectories>
+      <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
```
MikhailArkhipov: Out of curiosity, lack of entry means DLL and if entry is present it is static?

MikhailArkhipov:
:+1: 

Can be verified on clean VM by just launching the exe right (i.e. without anything else from VS installed)

src/Microsoft.R.Host.vcxproj
int19h: Lack of entry means default value, which is "Debug multithreaded DLL" for Debug, and "Release multithreaded DLL" for Release. So any other option requires explicitly setting the property. Available options are all combinations of debug/release, single-threaded/multithreaded, and DLL/static.

int19h:
I used depends.exe to inspect the link table of the binary and see which DLLs it uses. Before the change:

image

after the change:

image

But yes, verification on a clean VM would be a nice confirmation.

PR #70, opened 2016-07-01
Host can send binary blobs.

karthiknadig:
Host can send binary blobs. Plots updated to use blobs. Added =r (raw…) eval kind. First element in the message json now has the header [id, name, blob_count, <optional>request_id].

see https://github.com/Microsoft/RTVS/issues/2023

### DIFF
--- a/src/blobs.cpp
+++ b/src/blobs.cpp
@@ -1,0 +1,85 @@
+
+/* ****************************************************************************
+*
+* Copyright (c) Microsoft Corporation. All rights reserved.
+*
+*
+* This file is part of Microsoft R Host.
+*
+* Microsoft R Host is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 2 of the License, or
+* (at your option) any later version.
+*
+* Microsoft R Host is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with Microsoft R Host.  If not, see <http://www.gnu.org/licenses/>.
+*
+* ***************************************************************************/
+
+#include "blobs.h"
+#include "log.h"
+
+using namespace rhost::util;
+using namespace rhost::log;
+namespace js = picojson;
+
+namespace rhost {
+    namespace raw {
+        namespace {
+            void blob_error(SEXP sexp, const char* format, ...) {
+                SEXP repr_sexp = STRING_ELT(Rf_deparse1line(sexp, R_FALSE), 0);
+                Rf_protect(repr_sexp);
+                const char* repr = R_CHAR(repr_sexp);
+
+                char buf[0x10000] = {};
+                snprintf(buf, sizeof buf, "BLOB serialization failed for input:\n\n%s\n\n", repr);
+
+                size_t len = strlen(buf);
+                va_list va;
+                va_start(va, format);
+                vsnprintf(buf + len, sizeof buf - len, format, va);
+                va_end(format);
+
+                Rf_error("%s", buf);
+            }
+
+            void at_most_one(SEXP sexp) {
+                if (Rf_length(sexp) != 1) {
+                    blob_error(sexp, "Vector must have 0 or 1 elements.");
+                }
+            }
+
+            bool to_blobs_internal(SEXP sexp, rhost::util::blobs& blobs, picojson::value& json) {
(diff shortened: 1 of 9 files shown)


src/blobs.cpp
```diff
@@ -0,0 +1,85 @@
...
+                size_t len = strlen(buf);
+                va_list va;
+                va_start(va, format);
+                vsnprintf(buf + len, sizeof buf - len, format, va);
+                va_end(format);
+
+                Rf_error("%s", buf);
+            }
+
+            void at_most_one(SEXP sexp) {
```
int19h: Unused?

src/blobs.cpp
```diff
@@ -0,0 +1,85 @@
...
+                    rhost::util::append(blobs, blob(data, length));
+                    return true;
+                }
+
+                blob_error(sexp, "Unsupported type for blob - must be one of: NULL; raw.");
+                return false;
+            }
+        }
+
+        bool to_blobs(SEXP sexp, rhost::util::blobs& blobs, js::value& json) {
```
int19h: `json` is effectively unused, and looks like it would be unused here in all future cases too?

src/blobs.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
```
int19h: Duplicate (there's one below the comment).

src/grdeviceside.cpp
```diff
@@ -708,12 +708,35 @@ namespace rhost {
             void ide_device::send(const std::tr2::sys::path& filename) {
                 auto path_copy(filename);
                 rhost::host::with_cancellation([&] {
+                    std::string file_path = path_copy.make_preferred().string();
+                    rhost::util::blobs data;
+                    char* plot_image_data = nullptr;
```
int19h: `std::unique_ptr<char[]>`

src/grdeviceside.cpp
```diff
@@ -708,12 +708,35 @@ namespace rhost {
...
+
+                    if (file_path.length() > 0) {
+                        // Send actual plot image as Binary
+                        std::ifstream plot_file_stream(file_path, std::ios::binary | std::ios::ate);
+                        std::ifstream::pos_type file_end = plot_file_stream.tellg();
+                        plot_image_data = new char[file_end];
+                        plot_file_stream.seekg(0, std::ios::beg);
+                        plot_file_stream.read(plot_image_data, file_end);
+                        size = static_cast<size_t>(file_end);
+                        plot_file_stream.close();
```
int19h: Destructor will close it when it goes out of scope.

src/util.h
```diff
@@ -117,6 +117,20 @@ namespace rhost {
             }
         };
 
+        struct blob {
```
int19h: This is effectively a view of a slice of a blob, rather than a blob (i.e. it doesn't manage the lifetime of the data, and is not valid as soon as the buffer into which `raw_data` points goes away). It probably deserves a more descriptive name that would make it clear - something like `blob_slice` - and some comments to that effect.

int19h: Also, consider how this will be used (if it will be) for receiving messages with blobs. We'll probably need a separate type there that would actually hold data / manage lifetime.

src/util.h
```diff
@@ -117,6 +117,20 @@ namespace rhost {
...
+            void* raw_data;
```
int19h: For a view, this should probably be `const void`, since we don't actually mutate the data anywhere, and sender might have an immutable buffer to send from (say, `const std::string`).

src/util.h
```diff
@@ -117,6 +117,20 @@ namespace rhost {
...
+            size_t length;
+
+            blob() :
+                raw_data(nullptr), length(0) {}
+            blob(void* const data, const size_t size) :
```
int19h: `const` on arguments doesn't seem to be doing anything useful here.

src/util.h
```diff
@@ -117,19 +117,14 @@ namespace rhost {
...
-                raw_data(data), length(size) {}
-
-            inline bool is_empty() { return (!raw_data) || (!length); }
-        };
-
-        typedef std::vector<blob> blobs;
+        inline void append_from_file(blob& blob, const std::string& path) {
+            std::ifstream ifs(path, std::ios::binary | std::ios::in);
+            std::noskipws(ifs);
+            blob.push_back(blob_slice(std::istream_iterator<byte>(ifs), std::istream_iterator<byte>()));
```
int19h: This will actually read it byte by byte, so it'll be super slow (~2 orders of magnitude).

I think it's best to ditch iostreams here - they just aren't a good fit for binary data. Might as well just `fopen` and read the whole thing as a single chunk. `SCOPE_WARDEN` can be used to automatically clean up.

int19h:
:+1:

int19h:
:+1:

PR #72, opened 2016-07-14
Adds feature to create and send binary blobs to RHost 

karthiknadig:
This change has following features related to blobs : [2065](https://github.com/Microsoft/RTVS/issues/2065)
1. RHost now supports receiving blobs in batches.
2. Adds protocol changes to support creation, access, and deletion of blobs in batches.
3. Blobs can be created, accessed, and destroyed from R

### DIFF
--- a/src/host.cpp
+++ b/src/host.cpp
@@ -313,8 +313,13 @@ namespace rhost {
                 return data;
             }
 
-            for (blob_slice& slice : (*res).second) {
-                data.reserve(data.size() + slice.size());
+            size_t totalSize = 0;
+            for (blob_slice& slice : res->second) {
+                totalSize += slice.size();
+            }
+
+            data.reserve(totalSize);
+            for (blob_slice& slice : res->second) {
                 data.insert(data.end(), slice.begin(), slice.end());
             }
 
@@ -924,17 +929,17 @@ namespace rhost {
         }
 
         void handle_binary_message(ws_connection_type::message_ptr msg) {
+            std::lock_guard<std::mutex> lock(binary_message_queue_mutex);
+
             if (binary_message_queue.size() != 1) {
-                fatal_error("Unexpected incoming client message - host was expecting a single pending binary message.");
+                fatal_error("Unexpected incoming client message - host was expecting a binary message.");
             }
 
             message incoming = binary_message_queue.front();
             const auto& raw = msg->get_raw_payload();
-            rhost::util::blob_slice slice(raw.data(), raw.data() + raw.size());
-            incoming.blob.push_back(std::move(slice));
+            incoming.blob.emplace_back(rhost::util::blob_slice(raw.data(), raw.data() + raw.size()));
 
             if (incoming.blob_count == incoming.blob.size()) {
-                std::lock_guard<std::mutex> lock(binary_message_queue_mutex);
                 create_blob(incoming);
                 binary_message_queue.pop();
              }
--- a/src/host.cpp
+++ b/src/host.cpp
@@ -122,6 +122,15 @@ namespace rhost {
         std::string eval_cancel_target; // ID of the eval on the stack that is the cancellation target
         std::mutex eval_stack_mutex;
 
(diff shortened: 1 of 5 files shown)


src/host.cpp
```diff
@@ -274,6 +283,75 @@ namespace rhost {
...
+
+        const blob_slice get_blob_as_single_slice(int blob_id) {
+            std::map<int, rhost::util::blob>::iterator res = received_blobs.find(blob_id);
+            blob_slice data;
+
+            if (res == received_blobs.end()) {
+                return data;
+            }
+
+            for (blob_slice& slice : (*res).second) {
```
int19h: `->` instead of `(*...).`

src/host.cpp
```diff
@@ -274,6 +283,75 @@ namespace rhost {
...
+                data.reserve(data.size() + slice.size());
```
int19h: It seems like there could be a separate loop to compute the size first, and reserve the whole thing, before the loop that actually does the copying, to avoid any reallocs at all.

src/host.cpp
```diff
@@ -838,6 +923,34 @@ namespace rhost {
...
 
+        void handle_binary_message(ws_connection_type::message_ptr msg) {
+            if (binary_message_queue.size() != 1) {
+                fatal_error("Unexpected incoming client message - host was expecting a single pending binary message.");
+            }
+
+            message incoming = binary_message_queue.front();
+            const auto& raw = msg->get_raw_payload();
+            rhost::util::blob_slice slice(raw.data(), raw.data() + raw.size());
+            incoming.blob.push_back(std::move(slice));
```
int19h: This will be even faster and shorter with `emplace_back` to construct the slice directly inside the vector.

src/host.cpp
```diff
@@ -838,6 +923,34 @@ namespace rhost {
             unblock_message_loop();
         }
 
+        void handle_binary_message(ws_connection_type::message_ptr msg) {
+            if (binary_message_queue.size() != 1) {
+                fatal_error("Unexpected incoming client message - host was expecting a single pending binary message.");
+            }
+
+            message incoming = binary_message_queue.front();
```
int19h: Looks like this check should also be under queue mutex, since it reads something that can be mutated from another thread.

src/r_util.cpp
```diff
@@ -441,6 +441,42 @@ namespace rhost {
...
+
+        extern "C" SEXP get_blob(SEXP id) {
+            int blob_id = Rf_asInteger(id);
+            const std::vector<byte> data = rhost::host::get_blob_as_single_slice(blob_id);
+
+            if (data.size() == 0) {
+                return R_NilValue;
+            }
+
+            SEXP rawVector = Rf_allocVector(RAWSXP, data.size());
```
int19h: This probably needs to be `Rf_protect`-ed - I don't think `RAW` can potentially trigger a GC, but I don't think it guarantees it, either.

src/host.cpp
```diff
@@ -924,17 +929,17 @@ namespace rhost {
...
             if (binary_message_queue.size() != 1) {
-                fatal_error("Unexpected incoming client message - host was expecting a single pending binary message.");
+                fatal_error("Unexpected incoming client message - host was expecting a binary message.");
             }
 
             message incoming = binary_message_queue.front();
             const auto& raw = msg->get_raw_payload();
-            rhost::util::blob_slice slice(raw.data(), raw.data() + raw.size());
-            incoming.blob.push_back(std::move(slice));
+            incoming.blob.emplace_back(rhost::util::blob_slice(raw.data(), raw.data() + raw.size()));
```
int19h: You don't need to invoke the constructor explicitly with `emplace` - it will automatically invoke it, and forward all the arguments. So you can just do:
```cpp
emplace_back(raw.data(), raw.data() + raw.size())
```

int19h:
:+1:

int19h:
:+1:

PR #73, opened 2016-07-26
Optimize protocol for better broker performance [R-Host]

int19h:
https://github.com/Microsoft/RTVS/issues/2082

Switch protocol to use a single binary websocket packet for every message, to allow message header to be inspected without the overhead JSON parsing.

### DIFF
--- a/src/Microsoft.R.Host.vcxproj
+++ b/src/Microsoft.R.Host.vcxproj
@@ -102,6 +102,7 @@
       <SDLCheck>false</SDLCheck>
       <AdditionalIncludeDirectories>..\lib\picojson;..\lib\websocketpp</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
+      <DisableSpecificWarnings>4200</DisableSpecificWarnings>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
@@ -120,6 +121,7 @@
       <SDLCheck>false</SDLCheck>
       <AdditionalIncludeDirectories>..\lib\picojson;..\lib\websocketpp</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
+      <DisableSpecificWarnings>4200</DisableSpecificWarnings>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
@@ -140,6 +142,7 @@
       <SDLCheck>false</SDLCheck>
       <AdditionalIncludeDirectories>..\lib\picojson;..\lib\websocketpp</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
+      <DisableSpecificWarnings>4200</DisableSpecificWarnings>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
@@ -162,6 +165,7 @@
       <SDLCheck>false</SDLCheck>
       <AdditionalIncludeDirectories>..\lib\picojson;..\lib\websocketpp</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
+      <DisableSpecificWarnings>4200</DisableSpecificWarnings>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
@@ -178,6 +182,7 @@
     <ClCompile Include="exports.cpp" />
     <ClCompile Include="grdeviceside.cpp" />
     <ClCompile Include="grdevicesxaml.cpp" />
+    <ClCompile Include="message.cpp" />
     <ClCompile Include="msvcrt.cpp" />
     <ClCompile Include="eval.cpp" />
     <ClCompile Include="log.cpp" />
@@ -195,6 +200,7 @@
     <ClInclude Include="grdevices.h" />
     <ClInclude Include="grdeviceside.h" />
     <ClInclude Include="grdevicesxaml.h" />
+    <ClInclude Include="message.h" />
     <ClInclude Include="msvcrt.h" />
     <ClInclude Include="eval.h" />
     <ClInclude Include="log.h" />
--- a/src/blobs.cpp
+++ b/src/blobs.cpp
@@ -28,7 +28,7 @@ using namespace rhost::log;
 namespace js = picojson;
 
 namespace rhost {
-    namespace raw {
+    namespace blobs {
         namespace {
             void blob_error(SEXP sexp, const char* format, ...) {
                 SEXP repr_sexp = STRING_ELT(Rf_deparse1line(sexp, R_FALSE), 0);
@@ -47,17 +47,17 @@ namespace rhost {
                 Rf_error("%s", buf);
             }
 
-            bool to_blob_internal(SEXP sexp, rhost::util::blob& blob) {
+            bool to_blob_internal(SEXP sexp, std::vector<char>& blob) {
                 int type = TYPEOF(sexp);
                 size_t length = Rf_length(sexp);
 
-                if ((type == NILSXP) || (length == 0)) {
-                    return true;
+                if (type == NILSXP) {
+                    return false;
                 }
 
                 if (type == RAWSXP) {
                     Rbyte* data = RAW(sexp);
-                    blob.push_back(rhost::util::blob_slice(data, data + length));
+                    blob.assign(data, data + length);
                     return true;
                 }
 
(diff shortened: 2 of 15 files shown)


int19h:
Corresponding RTVS change:
https://github.com/Microsoft/RTVS/pull/2116

src/Microsoft.R.Host.vcxproj
```diff
@@ -102,6 +102,7 @@
       <SDLCheck>false</SDLCheck>
       <AdditionalIncludeDirectories>..\lib\picojson;..\lib\websocketpp</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
+      <DisableSpecificWarnings>4200</DisableSpecificWarnings>
```
int19h: This is for 0-length array in `message_repr`. It's technically non-standard C++, but VC++, gcc and Clang all support it these days, and it makes for more readable code than otherwise.

int19h:
Ping.

src/blobs.cpp
```diff
@@ -66,10 +66,36 @@ namespace rhost {
...
         }
 
-        bool to_blob(SEXP sexp, rhost::util::blob& blob) {
+        bool to_blob(SEXP sexp, std::vector<char>& blob) {
             bool result = false;
             rhost::util::errors_to_exceptions([&] {result = to_blob_internal(sexp, blob); });
             return result;
         }
+
+        void append_from_file(blob& blob, const char* path) {
```
MikhailArkhipov: Should path be Unicode? Or converted from UTF-8?

src/host.cpp
```diff
@@ -145,108 +142,83 @@ namespace rhost {
...
+
+            auto& repr = *reinterpret_cast<message_repr*>(buf.get());
+            repr.id = id;
+            repr.request_id = request_id;
+
+            char* p = repr.data;
 
-            picojson::value value(picojson::array_type, false);
-            auto& array = value.get<picojson::array>();
+            strcpy(p, name.c_str());
```
MikhailArkhipov: memcpy may be a tiny bit faster

src/host.cpp
```diff
@@ -284,101 +256,98 @@ namespace rhost {
...
+                fatal_error("Blob ID overflow");
             }
-        }
 
-        int create_blob(const rhost::util::blob& blob) {
-            std::lock_guard<std::mutex> lock(received_blobs_mutex);
-            int blob_id = ++next_blob_id;
-            received_blobs[blob_id] = blob;
-            return blob_id;
+            blobs[id] = std::move(blob);
```
MikhailArkhipov: should it be under lock or is assignment atomic

src/host.cpp
```diff
@@ -284,101 +256,98 @@ namespace rhost {
...
+            std::lock_guard<std::mutex> lock(blobs_mutex);
+            blob_id id = ++next_blob_id;
 
-            {
-                std::lock_guard<std::mutex> lock(received_blobs_mutex);
-                int blob_id = ++next_blob_id;
-                received_blobs[blob_id] = msg.blob;
-                respond_to_message(msg, static_cast<double>(blob_id));
+            // Check that it never overflows double mantissa, and provide immediate diagnostics if it happens.
+            if (id != blob_id(double(id))) {
```
MikhailArkhipov: Out or curiosity, do you think 64-bit it would overflow? Or there is no portable type?

src/host.cpp
```diff
@@ -284,101 +256,98 @@ namespace rhost {
...
-                blob_slice slice = get_blob_as_single_slice(blob_id);
-                if (slice.size() > 0) {
-                    blob.push_back(slice);
-                    rhost::util::append(blob_ids, static_cast<double>(blob_id));
-                }
+            auto json = msg.json();
+            if (!json[0].is<double>()) {
+                fatal_error("GetBlob: non-numeric blob ID");
+            }
+            auto id = static_cast<blob_id>(json[0].get<double>());
```
MikhailArkhipov: Might be useful to have typedef BLOB_ID double and write .get<BLOB_ID>()

src/host.cpp
```diff
@@ -644,21 +614,21 @@ namespace rhost {
...
+                            msg.request_id(), msg.name(), id, name.c_str());
+                    } else if (strcmp(msg.name() + 1, name.c_str() + 1) != 0) {
+                        fatal_error("Response to [%llu,'%s'] has mismatched name '%s'.",
+                            id, name.c_str(), msg.name());
                     }
                     return msg;
                 }
 
-                if (msg.name.size() >= 2 && msg.name[0] == '?' && msg.name[1] == '=') {
+                if (msg.name()[0] == '?' && msg.name()[1] == '=') {
```
MikhailArkhipov: Length is now guaranteed to be > 1 ?

src/blobs.cpp
int19h: This is not new code, it just got moved from one file to another. 

As far as encoding handling, it's whatever the system's non-multibyte FS encoding is. This would be `CP_ACP` on Windows, and whatever the current locale specifies as encoding (normally UTF-8) on Unix. It's also consistent with how we handle file paths elsewhere in host code (you can see the call site for this uses `make_preferred`, which will use the right encoding).

We might want to review this later, but it'd have to be a sweep over the entire code, not just here.

src/host.cpp
int19h: `memcpy` doesn't give you a free `\0` at the end tho :)

src/host.cpp
int19h: It is under lock - see `lock_guard` above.

src/host.cpp
int19h: The problem here is that we're casting it to `double` for round-tripping (because neither JSON nor R have 64-bit integer type), and `double` can only preserve 52 bits of value out of 64 (the rest are exponent and sign). The other problem is that when it happens, it's silent - you just lose precision. So I figured I'd add a check for that, just in case. Still highly unlikely that someone would create more than 2^52 blobs, so this is really more of a code-as-documentation.

src/host.cpp
int19h: There's already `blob_id`, so that would be something like `blob_json_id` then?

src/host.cpp
int19h: `name()` now returns `const char*`, and it's a null-terminated string. So this is guaranteed to work due to short circuiting behavior - there's always [0] to compare, and if the string is 0-length, it'll be `\0`, so first comparison fails, and second never happens. If first comparison succeeds, then [1] is guaranteed to be there (but is possibly `\0`). Etc.

src/log.cpp
```diff
@@ -235,7 +235,7 @@ namespace rhost {
                     msgbox_text += c;
                 }
                 
-                MessageBoxA(HWND_DESKTOP, msgbox_text.c_str(), "Microsoft R Host Process fatal error", MB_OK | MB_ICONERROR);
+                //MessageBoxA(HWND_DESKTOP, msgbox_text.c_str(), "Microsoft R Host Process fatal error", MB_OK | MB_ICONERROR);
```
int19h: I removed this because we detour all message boxes to VS, and it already receives an error message (and shows it to the user) through other channels. This was originally here for debugging the host standalone.

src/host.cpp
MikhailArkhipov: Can they be the same type?

src/host.cpp
MikhailArkhipov: It does if source is a string and getting copied as size+1 right? :-)

src/blobs.cpp
MikhailArkhipov: I am mostly worried about Users\[LOCAL USER NAME]\[LOCALIZED DOC FOLDER]. But I guess it is no different from invoking say, write.csv() into localized folder from REPL

src/host.cpp
int19h: I can use `double` everywhere, but I really don't want to, because it's not the true value domain (then everything would have to check for non-integer doubles and raise errors, or just ignore and round with potentially strange behavior). Double is really used as a 52-bit int for transport purposes here, and is due to transport limitations. Long-term, I'm hoping to get rid of the JSON issue by moving to MsgPack (similar structure, but binary serialization, and natively supports blobs, 64-bit ints etc). And there might be a better way to represent this in R as well, e.g. as a pair of 32-bit ints.

src/host.cpp
```diff
@@ -847,113 +818,42 @@ namespace rhost {
...
             }
 
-            auto args = array.begin() + 1;
-            incoming.args.assign(args, array.end());
-
-            if (incoming.name == "!/") {
+            std::string name = incoming.name();
+            if (name == "!End") {
+                terminate("Shutdown request received.");
+            } else if (name == "!/") {
```
MikhailArkhipov: Might be useful to have a map of function pointers eventually like MFC or ATL_COM_MAP used to do

MikhailArkhipov:
👍

src/blobs.cpp
int19h: Yep. This should work currently, provided that they don't use any Unicode letters in the name that cannot be mapped into their default codepage. If they do, we'll break (but, again, it's not really a new issue).

For future refactoring, to mitigate this, we should start using `std::filesystem` types everywhere throughout the code. As I recall, those are supposed to be platform-aware, and use the internal representation that can represent any path (e.g. UCS2 on Windows, UTF-8 on Mac, native locale on Linux etc). I'll file a separate issue.

src/host.cpp
int19h: Yeah. I want to do a big refactoring of this code, splitting it into transport (just handles generic messages and invokes handlers via map) and actual host code. I actually started this change that way, but it's too big for this milestone at this point.

PR #74, opened 2016-07-27
Remote package lock state

karthiknadig:
Added code to determine lock state for a given dll. [2103](https://github.com/Microsoft/RTVS/issues/2103)

### DIFF
--- a/src/r_util.cpp
+++ b/src/r_util.cpp
@@ -475,12 +475,14 @@ namespace rhost {
         }
 
         extern "C" SEXP get_file_lock_state(SEXP paths) {
-            R_len_t len  = Rf_length(paths);
+            R_len_t len = Rf_length(paths);
             std::vector<std::wstring> files;
-            for (R_len_t i = 0; i < len; ++i) {
-                const wchar_t* file_path = Rf_wtransChar(STRING_ELT(paths, i));
-                files.emplace_back(file_path);
-            }
+            r_top_level_exec([&]() {
+                for (R_len_t i = 0; i < len; ++i) {
+                    const wchar_t* file_path = Rf_wtransChar(STRING_ELT(paths, i));
+                    files.emplace_back(file_path);
+                }
+            }, __FUNCTION__);
             file_lock_state lock_state = lock_state_by_file(files);
             return Rf_ScalarInteger(static_cast<int>(lock_state));
         }
--- a/src/r_util.cpp
+++ b/src/r_util.cpp
@@ -474,10 +474,16 @@ namespace rhost {
             return R_NilValue;
         }
 
-        extern "C" SEXP get_package_lock_state(SEXP path) {
-            const char* file_path = R_CHAR(STRING_ELT(path, 0));
-            SEXP pkg_lock_state = Rf_mkCharCE(rhost::util::lock_state_by_file(file_path), CE_UTF8);
-            return Rf_ScalarString(pkg_lock_state);
+        extern "C" SEXP get_file_lock_state(SEXP paths) {
+            R_len_t len  = Rf_length(paths);
+            std::vector<std::string> files;
+            
+            for (R_len_t i = 0; i < len; ++i) {
+                const char* file_path = R_CHAR(STRING_ELT(paths, i));
+                files.emplace_back(file_path);
+            }
+            file_lock_state lock_state = lock_state_by_file(files);
+            return Rf_ScalarInteger(static_cast<int>(lock_state));
         }
 
         R_CallMethodDef call_methods[] = {
@@ -495,7 +501,7 @@ namespace rhost {
             { "Microsoft.R.Host::Call.create_blob", (DL_FUNC)create_blob, 1 },
             { "Microsoft.R.Host::Call.get_blob", (DL_FUNC)get_blob, 1 },
             { "Microsoft.R.Host::Call.destroy_blob", (DL_FUNC)destroy_blob, 1 },
-            { "Microsoft.R.Host::Call.get_package_lock_state", (DL_FUNC)get_package_lock_state, 1 },
+            { "Microsoft.R.Host::Call.get_file_lock_state", (DL_FUNC)get_file_lock_state, 1 },
             { }
         };
 
--- a/src/r_util.cpp
+++ b/src/r_util.cpp
@@ -28,6 +28,7 @@
 #include "host.h"
 #include "json.h"
 #include "exports.h"
+#include "rstrtmgr.h"
 
 using namespace rhost::log;
 using namespace rhost::util;
@@ -473,6 +474,12 @@ namespace rhost {
             return R_NilValue;
         }
 
+        extern "C" SEXP get_package_lock_state(SEXP path) {
+            const char* file_path = R_CHAR(STRING_ELT(path, 0));
+            SEXP pkg_lock_state = Rf_mkCharCE(rhost::util::lock_state_by_file(file_path), CE_UTF8);
+            return Rf_ScalarString(pkg_lock_state);
+        }
+
         R_CallMethodDef call_methods[] = {
             { "Microsoft.R.Host::Call.unevaluated_promise", (DL_FUNC)unevaluated_promise, 2 },
             { "Microsoft.R.Host::Call.memory_connection", (DL_FUNC)memory_connection_new, 4 },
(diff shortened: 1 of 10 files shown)


src/rstrtmgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+            auto err = mbstowcs_s(&wchar_len, fpath, file_path, char_len);
+            if (err != 0) {
+                Rf_error("Unable to check the library lock state.");
+            }
+
+            std::unique_ptr<WCHAR> session_uuid_str(new WCHAR[128]);
+            DWORD session_handle = 0;
+            DWORD error = RmStartSession(&session_handle, 0, session_uuid_str.get());
+            if (error == ERROR_SUCCESS) {
```
int19h: Should the code even try to continue (with the loop below) if this fails? It seems that it should rather check for error and return early.

src/rstrtmgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+                        RM_PROCESS_INFO* pinfo = (process_info_data.get() + i);
+                        if (pinfo != nullptr) {
+                            process_ids.push_back(pinfo->Process.dwProcessId);
+                        }
+                    }
+                    break;
+                }
+            } while (error == ERROR_MORE_DATA);
+
+            RmEndSession(session_handle);
```
int19h: `SCOPE_WARDEN`

src/rstrtmgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "Rapi.h"
+#include <RestartManager.h>
+
+namespace rhost {
+    namespace util {
+#define PACKAGE_LOCK_STATE_UNLOCKED        "unlocked"
+#define PACKAGE_LOCK_STATE_LOCKED_BY_R     "locked_by_r"
+#define PACKAGE_LOCK_STATE_LOCKED_BY_OTHER "locked_by_other"
+
+        const char* lock_state_by_file(const char* file_path) {
```
int19h: Since this is a C API, it should return an `enum` value. The R wrapper should then convert those to the corresponding R value. Also, have a look at `factor()` in R - I think that it's probably the best way to map enums.

src/rstrtmgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+            UINT process_info_needed = 0;
+            UINT process_info_count = 0;
+            DWORD reboot_reason = RmRebootReasonNone;
+            if (error == ERROR_SUCCESS) {
+                error = RmGetList(session_handle, &process_info_needed, &process_info_count, nullptr, &reboot_reason);
+            }
+
+            std::vector<DWORD> process_ids;
+            do {
+                std::unique_ptr<RM_PROCESS_INFO> process_info_data(new RM_PROCESS_INFO[process_info_needed]);
```
int19h: This needs to be `unique_ptr<RM_PROCESS_INFO[]>` (otherwise you're mismatching array `new[]` with non-array `delete`). But `vector` is better still.

Also, either `unique_ptr<...[]>` or `vector` will overload `[i]`, so you can use that below instead of `get() + i`.

src/rstrtmgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+                process_info_count = process_info_needed;
+                error = RmGetList(session_handle, &process_info_needed, &process_info_count, process_info_data.get(), &reboot_reason);
+                if (error == ERROR_SUCCESS) {
+                    for (UINT i = 0; i < process_info_count; ++i) {
+                        RM_PROCESS_INFO* pinfo = (process_info_data.get() + i);
+                        if (pinfo != nullptr) {
```
int19h: Redundant check - this pointer can never be null.

src/rstrtmgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+                error = RmGetList(session_handle, &process_info_needed, &process_info_count, nullptr, &reboot_reason);
+            }
+
+            std::vector<DWORD> process_ids;
+            do {
+                std::unique_ptr<RM_PROCESS_INFO> process_info_data(new RM_PROCESS_INFO[process_info_needed]);
+                process_info_count = process_info_needed;
+                error = RmGetList(session_handle, &process_info_needed, &process_info_count, process_info_data.get(), &reboot_reason);
+                if (error == ERROR_SUCCESS) {
+                    for (UINT i = 0; i < process_info_count; ++i) {
```
int19h: Once the check below goes away, this entire loop looks like it can be replaced with `std::transform`.

src/rstrtmgr.cpp
```diff
@@ -0,0 +1,86 @@
...
+*
+* ***************************************************************************/
+
+#include "rstrtmgr.h"
+#include "Rapi.h"
+#include <RestartManager.h>
+
+namespace rhost {
+    namespace util {
+#define PACKAGE_LOCK_STATE_UNLOCKED        "unlocked"
```
int19h: #define should only be used for macros (copy/pasted stuff from R headers excepted). These are best declared as `const char ...[] = ...`.

src/r_util.cpp
```diff
@@ -478,6 +479,12 @@ namespace rhost {
             return R_NilValue;
         }
 
+        extern "C" SEXP get_package_lock_state(SEXP path) {
```
int19h: This appears to be a generic function to check whether any given file (not just package) is locked - the name should reflect that.

src/r_util.cpp
```diff
@@ -473,6 +474,18 @@ namespace rhost {
             return R_NilValue;
         }
 
+        extern "C" SEXP get_file_lock_state(SEXP paths) {
+            R_len_t len  = Rf_length(paths);
+            std::vector<std::string> files;
+            
+            for (R_len_t i = 0; i < len; ++i) {
+                const char* file_path = R_CHAR(STRING_ELT(paths, i));
```
int19h: This will get the string data in whatever encoding it was stored in the R character vector, which can vary. It will _usually_ be CP_ACP, but can also be Latin-1 or UTF-8. You need to call `Rf_translateCharUTF8`.

This will add another problem, in that `Rf_translateCharUTF8` can raise an R error, which will unwind the stack using `setjmp` (skipping any destructors). Meaning that you either have to avoid allocating any C++ objects on the stack before you call it, or you need to use `r_top_level_exec`.

src/util.cpp
```diff
@@ -92,12 +92,28 @@ namespace rhost {
...
+            // Convert 8-bit characters to Unicode via Windows CP. This guarantees
+            // if locale for non-Unicode programs is set correctly, user can type in
+            // their language. This does NOT guarantee that all languages can be used
+            // since R is not Unicode app. If host app is Unicode, it must perform
+            // checks if text being passed here is convertible to Unicode.
+            std::wstring ws;
+            size_t cch = strlen(buf);
+            if (cch > 0) {
+                ws.resize(cch);
+                RString2Unicode(&ws[0], (char*)buf, len);
```
int19h: This is only useful for a limited subset of strings returned by R (where it uses that funky ASCII-with-embedded-UTF encoding). If I remember correctly, that's for strings that are passed to callbacks like `WriteConsole` etc. Anyway, if you get UTF-8 from R, you can use `std::codecvt_utf8` to get wide chars in a standard way.

src/rstrtmgr.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+            UINT process_info_needed = 0;
+            UINT process_info_count = 0;
+            DWORD reboot_reason = RmRebootReasonNone;
+            error = RmGetList(session_handle, &process_info_needed, &process_info_count, nullptr, &reboot_reason);
+            if (error != ERROR_SUCCESS && error != ERROR_MORE_DATA) {
+                return file_lock_state::unlocked;
+            }
+
+            std::vector<DWORD> process_ids;
```
int19h: Looks like this isn't used inside the loop that follows - so it can be declared after the loop, and get the size directly when constructed (instead of potentially being resized several times in the loop).

src/util.cpp
int19h: Actually, I see that there's already code to do UTF-to-wchar in `from_utf8` - it does it as the very first step. You can just refactor that out into a function.

int19h: Also, there's a function called `strsxp_to_utf8` in json.cpp, that I think is more in spirit of what you're trying to do... we just need an equivalent that returns `wstring` rather than `string`. They could both live in util.h.

src/r_util.cpp
```diff
@@ -476,10 +476,9 @@ namespace rhost {
 
         extern "C" SEXP get_file_lock_state(SEXP paths) {
             R_len_t len  = Rf_length(paths);
-            std::vector<std::string> files;
-            
+            std::vector<std::wstring> files;
             for (R_len_t i = 0; i < len; ++i) {
-                const char* file_path = R_CHAR(STRING_ELT(paths, i));
+                const wchar_t* file_path = Rf_wtransChar(STRING_ELT(paths, i));
```
int19h: Needs `r_top_level_exec` around the call to `Rf_wtransChar` to make sure that R errors from the latter don't cause memory leaks.

int19h:
:+1:

PR #75, opened 2016-07-28
Update websocketpp and picojson submodules to newer versions.

int19h:
For websocketpp, mostly a bunch of bugfixes.

For picojson, bugfixes (including a fix to the problem that forced us to disable some checks in debug builds because it made things too slow), and improved C++11 support - move constructors etc.


huguesv:
👍 No changes needed in r-host?  If we did disable some checks in debug build, we could re-enable them perhaps?

karthiknadig:
:+1:

PR #76, opened 2016-07-30
Register fail handler for websocket connection

int19h:
So that connection failures actually get reported and do not hang the host.


MikhailArkhipov:
:+1:

PR #77, opened 2016-08-08
Added function to fetch file from remote RHost.

karthiknadig:
Implements fetch feature. R user will be able to fetch any file they have access on RHost back to RTVS.

### DIFF
--- a/src/r_util.cpp
+++ b/src/r_util.cpp
@@ -34,6 +34,7 @@ using namespace rhost::log;
 using namespace rhost::util;
 using namespace rhost::host;
 using namespace rhost::json;
+namespace fs = std::experimental::filesystem;
 
 namespace rhost {
     namespace r_util {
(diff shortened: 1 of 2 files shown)


int19h:
:+1:

src/r_util.cpp
```diff
@@ -487,6 +488,22 @@ namespace rhost {
...
+        extern "C" SEXP fetch_file(SEXP path) {
+            fs::path fpath;
+            r_top_level_exec([&]() {
+                fpath = fs::u8path(Rf_translateCharUTF8(STRING_ELT(path, 0)));
+            }, __FUNCTION__);
+
+            if (!fpath.empty()) {
+                blobs::blob file_data;
+                blobs::append_from_file(file_data, fpath);
+                host::send_notification("!FetchFile", file_data, fpath.filename().string());
```
int19h: Given that we have binary evals now, would it be better to make this return the blob directly, instead of sending a custom notification message?

karthiknadig: Can you elaborate? They way I see it, this is no different from plot. User invokes a function, and RTVS has to handle it in a unique manner. The contents of the notification, in this case have to be saved. 

---
In reply to: [73976559](https://github.com/Microsoft/R-Host/pull/77#discussion_r73976559) [](ancestors = 73976559)

int19h: The reason why plot needs its own notification is because it's triggered indirectly by the user, not directly by us. In this case, you're invoking this function directly via eval, so you might as well get its return value directly, too, using `REvaluationKind.RawResult`. Unless I'm missing something?

karthiknadig: No, this is not invoked via eval. Scenario is: User runs some R code that writes to some file (on the remote machine). User wants to open the file locally. User executes fetch_file with the remote path.

since we don't have remote save file dialog, there is no way for us to know the path to fetch the file. so user will be triggering this.


---
In reply to: [74119967](https://github.com/Microsoft/R-Host/pull/77#discussion_r74119967) [](ancestors = 74119967)

int19h: Ah, now I see. All good then!

:+1:

PR #78, opened 2016-08-12
Adds features to support remote F5.

karthiknadig:


src/blobs.cpp
```diff
@@ -97,5 +99,28 @@ namespace rhost {
...
+
+        void save_to_file(blob_id id, const fs::path& file_path) {
+            auto data = rhost::host::get_blob(id);
+
+            fs::path parent = file_path.parent_path();
+            if (!fs::exists(parent)) {
+                fs::create_directories(parent);
+            }
+
+            FILE *f = std::fopen(file_path.string().c_str(), "wb");
```
karthiknadig: >file_path [](start = 33, length = 9)

file_path.make_preferred()

src/r_util.cpp
```diff
@@ -504,6 +506,37 @@ namespace rhost {
             return R_FalseValue;
         }
 
+        extern "C" SEXP save_project(SEXP id, SEXP project_name, SEXP dest_dir, SEXP temp_dir) {
```
AlexanderSher: There can be any set of files in the zip, not just project, correct?

src/r_util.cpp
```diff
@@ -504,6 +506,37 @@ namespace rhost {
...
+                prj_name = fs::u8path(Rf_translateCharUTF8(STRING_ELT(project_name, 0)));
+                d_dir = fs::u8path(Rf_translateCharUTF8(STRING_ELT(dest_dir, 0)));
+                t_dir = fs::u8path(Rf_translateCharUTF8(STRING_ELT(temp_dir, 0)));
+            }, __FUNCTION__);
+
+            fs::path zip_file = t_dir / prj_name;
+            zip_file.replace_extension(".zip");
+
+            if (fs::exists(zip_file)) {
+                fs::remove(zip_file);
```
AlexanderSher: It is possible that `zip_file` file got deleted after `fs::exists` is called

karthiknadig: That is fine. I have tested that case in my experiments. fs:remove returns false.

---
In reply to: [74645325](https://github.com/Microsoft/R-Host/pull/78#discussion_r74645325) [](ancestors = 74645325)

src/r_util.cpp
karthiknadig: This is intended for files under the project folder (under RTVS). So, 'yes' for any set of files as long as they are expected to be under the project folder.

---
In reply to: [74644583](https://github.com/Microsoft/R-Host/pull/78#discussion_r74644583) [](ancestors = 74644583)

src/r_util.cpp
AlexanderSher: Then there is no reason to call `fs::exists`

---
In reply to: [74651753](https://github.com/Microsoft/R-Host/pull/78#discussion_r74651753) [](ancestors = 74651753,74645325)

src/r_util.cpp
AlexanderSher: I'm just thinking if we should call it `save_to_project_folder` or something like this

---
In reply to: [74652029](https://github.com/Microsoft/R-Host/pull/78#discussion_r74652029) [](ancestors = 74652029,74644583)

src/r_util.cpp
```diff
@@ -521,6 +554,7 @@ namespace rhost {
             { "Microsoft.R.Host::Call.destroy_blob", (DL_FUNC)destroy_blob, 1 },
             { "Microsoft.R.Host::Call.get_file_lock_state", (DL_FUNC)get_file_lock_state, 1 },
             { "Microsoft.R.Host::Call.fetch_file", (DL_FUNC)fetch_file, 1 },
+            { "Microsoft.R.Host::Call.save_project", (DL_FUNC)save_project, 4 },
```
AlexanderSher: Do we want to provide any specific errors when this method fails?

src/project.cpp
```diff
@@ -0,0 +1,120 @@
...
+*
+* You should have received a copy of the GNU General Public License
+* along with Microsoft R Host.  If not, see <http://www.gnu.org/licenses/>.
+*
+* ***************************************************************************/
+
+
+#include "project.h"
+#include "util.h"
+#include <zip.h>
```
int19h: "" for this as well (<> is only for standard headers)

int19h: Also, can you move it to stdafx.h? Since it shouldn't be changing regularly on rebuilds...

src/project.cpp
```diff
@@ -0,0 +1,120 @@
...
+                if (zip_err != ZIP_ER_OK) {
+                    throw std::exception("Error while opening compressed project file.");
+                }
+
+                // Get the number of entries in the archive
+                zip_int64_t nfiles = zip_get_num_entries(archive, ZIP_FL_UNCHANGED);
+
+                // 10 MB buffer to extract files
+                size_t buff_size = 10 * 1024 * 1024;
+                std::unique_ptr<byte[]> buffer(new byte[buff_size]);
```
int19h: Where does `byte` come from? I suspect it's in Win32 headers somewhere, and we shouldn't depend on those where we don't have to. Just using `char` directly is fine - there's nothing useful that `typedef` actually adds here, since `sizeof(char)` is guaranteed to be 1.

src/project.cpp
```diff
@@ -0,0 +1,120 @@
...
+                    SCOPE_WARDEN(f_close, {
+                        if (f) {
+                            std::fclose(f);
+                        }
+                    });
+
+                    // Decompress the file
+                    zip_int64_t bytes_read = zip_fread(zf, buffer.get(), buff_size);
+                    while (bytes_read == buff_size) {
+                        std::fwrite(buffer.get(), sizeof(byte), static_cast<size_t>(bytes_read), f);
```
int19h: `fwrite` can fail - need to check return value.

(And if you swap the element size and element count arguments, you're guaranteed a return of 1 on success and 0 on failure, making this a tad more concise).

src/project.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+                    FILE* f = std::fopen(temp_file_name.make_preferred().string().c_str(), "wb");
+                    SCOPE_WARDEN(f_close, {
+                        if (f) {
+                            std::fclose(f);
+                        }
+                    });
+
+                    // Decompress the file
+                    zip_int64_t bytes_read = zip_fread(zf, buffer.get(), buff_size);
```
int19h: If the file is less than `buff_size`, we won't enter the loop below, and buffer will never be written out. Rewriting it as `do..while`, with read first and write following, should do the trick.

src/project.cpp
int19h: Ah, I see the second call to `fwrite` below. Still, it looks like it could be simplified to a single loop with a single read followed by a single write. If it happens to be zero-length, everything should still work fine.

src/project.cpp
```diff
@@ -0,0 +1,120 @@
...
+                // Copy all decompressed files to their destination
+                fs::copy(temp_dir, dest_dir, fs::copy_options::recursive | fs::copy_options::overwrite_existing);
+
+                // Delete all temp files created during decompression phase
+                fs::remove_all(temp_dir);
+            }
+        }
+
+        void extract_project(fs::path zip_file, fs::path dest_dir, fs::path temp_dir) {
+            return extract_project_internal(zip_file, dest_dir, temp_dir);
```
int19h: What's the purpose of separating `extract_project` and `extract_project_internal`?

src/r_util.cpp
```diff
@@ -504,6 +506,37 @@ namespace rhost {
...
+            blobs::save_to_file(blob_id, zip_file);
+
+            fs::path temp_proj_dir = t_dir / prj_name;
+            if (fs::exists(temp_proj_dir)) {
+                fs::remove_all(temp_proj_dir);
+            }
+
+            fs::path dest_proj_dir = d_dir / prj_name;
+
+            rproj::extract_project(zip_file, dest_proj_dir, temp_proj_dir);
```
int19h: This throws C++ exceptions on errors, but is intended to be called from R code. You will need to wrap this function using `exceptions_to_errors` helper, and only expose the wrapper function to R.

src/r_util.cpp
karthiknadig: @int19h Do you think we need to provide any additional error information?

---
In reply to: [74654678](https://github.com/Microsoft/R-Host/pull/78#discussion_r74654678) [](ancestors = 74654678)

src/r_util.cpp
int19h: Still not quite right. Same problem - there's a bunch of local stack-allocated variables here that have destructors that do meaningful cleanup (e.g. fs::path).

What you need is _all_ code that uses destructors to be inside extract_project, so that it can throw and unwind normally. Then wrap the whole enchilada with exceptions_to_errors. The function that calls exceptions_to_errors should not have any locals with non-trivial destructors.

---
In reply to: [74672717](https://github.com/Microsoft/R-Host/pull/78#discussion_r74672717) [](ancestors = 74672717)

int19h: Or maybe  not extract_project, since you also need to handle errors for save_to_file above. But extract it all to a separate function, e.g. save_to_project_folder_worker, and have save_to_project_folder call that from inside exceptions_to_errors.

---
In reply to: [74851442](https://github.com/Microsoft/R-Host/pull/78#discussion_r74851442) [](ancestors = 74851442,74672717)

src/r_util.cpp
int19h: I'm not sure how much we can provide even. Rf_error just takes an error message string, and exceptions_to_errors will automatically translate the exception message string to that.

---
In reply to: [74849277](https://github.com/Microsoft/R-Host/pull/78#discussion_r74849277) [](ancestors = 74849277,74654678)

src/r_util.cpp
```diff
@@ -489,19 +491,41 @@ namespace rhost {
...
+            });
+        }
+
+        extern "C" SEXP save_to_project_folder(SEXP id, SEXP project_name, SEXP dest_dir, SEXP temp_dir) {
+            auto blob_id = static_cast<blobs::blob_id>(Rf_asReal(id));
+            const char *prj_name = nullptr;
+            const char *t_dir = nullptr;
+            const char *d_dir = nullptr;
+
+            r_top_level_exec([&]() {
```
int19h: >r_top_level_exec [](start = 12, length = 16)

No need for this anymore - its only function is to swallow R errors (and let you examine them at leisure), but in this case you actually want R errors to propagate out of this function to the caller, if any Rf_translateChar calls fail.

int19h:
:shipit:

PR #79, opened 2016-08-15
Switch message transport from websockets to stdin/stdout.

int19h:
(for remoting - corresponding RTVS PR coming shortly)


src/main.cpp
```diff
@@ -214,6 +183,7 @@ namespace rhost {
 }
 
 int main(int argc, char** argv) {
+    //MessageBox(0, 0, 0, 0);
```
AlexanderSher: ?

int19h: Debug helper :) I'll get rid of it.

src/transport.cpp
```diff
@@ -0,0 +1,138 @@
...
+        void send_message(const message& msg) {
+            assert(output);
+
+            log_message("<==", msg.id(), msg.request_id(), msg.name(), msg.json_text(), msg.blob());
+
+            auto& payload = msg.payload();
+            boost::endian::little_uint32_buf_t msg_size(static_cast<uint32_t>(payload.size()));
+
+            std::lock_guard<std::mutex> lock(output_lock);
+            if (fwrite(&msg_size, sizeof msg_size, 1, output) == 1) {
```
AlexanderSher: What does `== 1` mean?

int19h: `fwrite` and `fread` return the number of elements read or written. In this case we're writing 1 element (3rd argument here) of size `sizeof msg_size` (2nd argument); so if write succeeded, the return value will be 1, and if it failed or partially succeeded, it'll be 0.

AlexanderSher:
:shipit:

karthiknadig:
:shipit:

PR #80, opened 2016-08-16
Fixes a crash in R-Host when running infinite recursive functions.

karthiknadig:
Fixes https://github.com/Microsoft/RTVS/issues/1973


int19h:
:+1:

PR #83, opened 2016-08-24
Support for multiple ide devices.

huguesv:
Add message for device creation to initialize new ide devices with the correct resolution, so that the first plot command on a new device renders to the exact size of the plot window, instead of rendering to default size and then having to resize immediately.
Add message for device destruction so that IDE can track lifetime of devices.
Add more info to the plot message, to identify the device it comes from as well as a unique plot id.
Add ability to remove any plot, not just the 'current'.
Add ability to copy a plot from a device to another.

### DIFF
--- a/src/grdeviceside.cpp
+++ b/src/grdeviceside.cpp
@@ -37,8 +37,10 @@ namespace rhost {
             class plot {
             public:
                 plot(pDevDesc dd);
+                plot(pDevDesc dd, plot* source_plot);
                 virtual ~plot();
 
+                boost::uuids::uuid get_id() const;
                 void set_pending_render();
                 bool has_pending_render() const;
                 bool pending_render_timeout_elapsed() const;
@@ -68,13 +70,15 @@ namespace rhost {
                 plot_history(pDevDesc dd);
 
                 plot* get_active() const;
+                plot* get_plot(const boost::uuids::uuid& plot_id);
+                bool select(const boost::uuids::uuid& plot_id);
                 void move_next();
                 void move_previous();
 
                 void new_page();
                 void append(std::unique_ptr<plot> plot);
                 void clear();
-                void remove_active();
+                void remove(const boost::uuids::uuid& plot_id);
 
                 void resize(double width, double height, double resolution);
                 void render_from_snapshot();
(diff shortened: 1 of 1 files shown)


src/grdeviceside.cpp
```diff
@@ -198,6 +206,13 @@ namespace rhost {
                 _has_pending_render(false) {
             }
 
+            plot::plot(pDevDesc dd, plot* source_plot) :
```
int19h: It looks like it's could be unified with the ctor above, with `source_plot` defaulting to `nullptr`.

src/grdeviceside.cpp
```diff
@@ -323,6 +342,30 @@ namespace rhost {
                 return (*_active_plot).get();
             }
 
+            plot* plot_history::get_plot(const boost::uuids::uuid& plot_id) {
+                for (auto it = _plots.begin(); it != _plots.end(); ++it) {
```
int19h: `for (auto&& plot : _plots)`

src/grdeviceside.cpp
```diff
@@ -323,6 +342,30 @@ namespace rhost {
...
+                }
+                return nullptr;
+            }
+
+            bool plot_history::select(const boost::uuids::uuid& plot_id) {
+                if (_active_plot != _plots.end() && (*_active_plot)->get_id() == plot_id) {
+                    return false;
+                }
+
+                for (auto it = _plots.begin(); it != _plots.end(); ++it) {
```
int19h: This looks like `std::find_if`.

src/grdeviceside.cpp
```diff
@@ -854,9 +982,30 @@ namespace rhost {
...
             static void process_pending_render(bool immediately) {
-                if (device_instance != nullptr) {
-                    device_instance->render_request(immediately);
+                for (auto dev : devices) {
+                    dev->render_request(immediately);
+                }
+            }
+
+            static ide_device* find_device_by_num(int device_num) {
+                for (auto dev : devices) {
```
int19h: Also looks like `find_if` would be appropriate here.

src/grdeviceside.cpp
```diff
@@ -870,94 +1019,267 @@ namespace rhost {
...
+                args = CDR(args);
+                SEXP param3 = CAR(args);
+
+                auto source_device_id = boost::lexical_cast<boost::uuids::uuid>(R_CHAR(STRING_ELT(param1, 0)));
+                auto source_plot_id = boost::lexical_cast<boost::uuids::uuid>(R_CHAR(STRING_ELT(param2, 0)));
+                auto target_device_id = boost::lexical_cast<boost::uuids::uuid>(R_CHAR(STRING_ELT(param3, 0)));
+
+                return rhost::util::exceptions_to_errors([&] {
+                    auto source_dev = find_device_by_id(source_device_id);
+                    if (source_dev != nullptr) {
```
int19h: The flow here is easier to follow if the nested `if`'s are unrolled, i.e.:
```c++
auto source_dev = ...;
if (source_dev == nullptr) { throw ... }
auto target_dev = ...;
if (source_dev == nullptr) { throw ... }
...
```

int19h:
:+1:

src/grdeviceside.cpp
huguesv: The thing is I need an iterator, because `_active_plot` that I want to set is an interator. 

`for (auto&& plot : _plots)` gets me a reference to a `unique_ptr`.

huguesv: I changed to use `find_if` in both `get_plot` and `select`.

int19h:
:+1:

PR #84, opened 2016-08-26
Fix #2290

int19h:
Fix https://github.com/Microsoft/RTVS/issues/2209
(Cannot decrease memory limit when running remote)

Do not try to change the memory limit if the computed limit is lower than the current value.


huguesv:
👍

PR #86, opened 2016-09-13
Set enable dpi awareness in manifest so that plots in the windows dev…

huguesv:
…ice aren't blurry.

https://github.com/Microsoft/RTVS/issues/2018


int19h:
:+1:, but let's keep an eye on anything breaking because of this (off the top of my head, someone might be trying to do something with Tk - although that, being very cross-plat, should be well-equipped to handle DPI differences).

PR #87, opened 2016-09-13
Fix 140/141 toolset

MikhailArkhipov:
Fix references to zlib (workaround until 15.0 targets arrive)


src/Microsoft.R.Host.vcxproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project DefaultTargets="Build" ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
MikhailArkhipov: Whoops, need to udo

PR #88, opened 2016-09-16
Blob stream implementation.

karthiknadig:
Addresses: https://github.com/Microsoft/RTVS/issues/2191

### DIFF
--- a/src/host.cpp
+++ b/src/host.cpp
@@ -339,7 +339,7 @@ namespace rhost {
             }
 
             blobs::blob::const_iterator begin = it->second.begin() + pos;
-            blobs::blob::const_iterator end = it->second.begin() + pos + count;
+            blobs::blob::const_iterator end = begin + count;
 
             blobs::blob part(begin, end);
             respond_to_message(msg, part);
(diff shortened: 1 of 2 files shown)


src/host.cpp
```diff
@@ -216,43 +227,143 @@ namespace rhost {
...
 
             std::lock_guard<std::mutex> lock(blobs_mutex);
             auto& it = blobs.find(id);
             if (it == blobs.end()) {
-                fatal_error("GetBlob: no blob with ID %llu", id);
+                fatal_error("GetBlobSize: no blob with ID %llu", id);
             }
 
-            respond_to_message(msg, it->second);
+            respond_to_message(msg, static_cast<double>(it->second.size()));
```
int19h: This will be 64-bit. Can you add the double overflow check here also? Might be time to put it into an inline function in util.h, something like `ensure_fits_double`.

src/host.cpp
```diff
@@ -216,43 +227,143 @@ namespace rhost {
...
-                    fatal_error("DestroyBlob: non-numeric blob ID");
-                }
+            auto& it = blobs.find(id);
+            if (it == blobs.end()) {
+                fatal_error("ReadBlob: no blob with ID %llu", id);
+            }
 
-                auto id = static_cast<blobs::blob_id>(val.get<double>());
-                blobs.erase(id);
+            if (((pos != 0) && pos >= static_cast<long long>(it->second.size())) || pos < 0) {
```
int19h: It looks like the `pos != 0` check here is redundant - the only time it'll be in play is when `pos` is 0, and `size()` is also zero, in which case empty blob is the correct response anyway.

int19h: Also, if we raise errors for negative `count`, why not for `pos`?

src/host.cpp
```diff
@@ -216,43 +227,143 @@ namespace rhost {
...
+            
+            // Read at position and count
+            size_t size = pos;
+            size += count;
+            if (count == -1 || size > it->second.size()) {
+                count = it->second.size() - pos;
+            }
+
+            blobs::blob::const_iterator begin = it->second.begin() + pos;
+            blobs::blob::const_iterator end = it->second.begin() + pos + count;
```
int19h: `begin + count`

src/host.cpp
```diff
@@ -216,43 +227,143 @@ namespace rhost {
...
+
+            blobs::blob part(begin, end);
+            respond_to_message(msg, part);
+        }
+
+        void write_blob(const message& msg) {
```
int19h: If I understand correctly, `ReadBlob` permits reads from arbitrary location, but `WriteBlob` doesn't? It feels like these should be consistent between each other (and if both allow offsets, you could expose it as a seekable stream).

src/host.cpp
karthiknadig: sure will do.

src/host.cpp
karthiknadig: I will update in the next iteration.

src/host.cpp
```diff
@@ -354,16 +354,35 @@ namespace rhost {
...
+                size += blob.size();
+                if (it->second.size() < size) {
+                    it->second.resize(size);
+                }
+
+                std::copy(blob.begin(), blob.end(), it->second.begin() + pos);
+            }
+            
             respond_to_message(msg, static_cast<double>(it->second.size()));
+            return;
```
int19h: Nit: redundant.

int19h:
👍

PR #89, opened 2016-09-20
Backport of multi plots support.

huguesv:


MikhailArkhipov:
If code matches what's in 0.6 then we are good. I used diff to make sure changes match.

MikhailArkhipov:
👍

PR #90, opened 2016-09-28
Logging changes

MikhailArkhipov:


src/log.h
```diff
@@ -25,32 +25,39 @@
 
 namespace rhost {
     namespace log {
+        enum class log_verbosity {
```
int19h: It kinda feels like this is the same as `log_level`, just named differently. Do we need these to be separate?

MikhailArkhipov: It is a valid question and usually verbosity matches level (i.e. errors are visible in minimal and so on). However, in our case verbosity more relates to privacy/security. So 'trace' may be OK as 'minimal' but error may not be logged at minimal if it contains PI or secure content. 

We may consider different name or even means to file error message without associated data

int19h: A different name sounds like a good idea then, for both of them - `log_level` should probably become `log_verbosity` to make it clear that it's just about how chatty we are, and `log_verbosity` should be `log_category`, and have specific enum levels for every category that we need to distinguish for security or PII purposes (i.e. no `minimal` or `full`, but more in the vein of `traffic` - e.g. `general_info` for things like version numbers or available memory, `user_input` for things like error messages that include expressions coming form the user, `user_data` for crash dumps etc).

int19h:
👍

src/log.h
MikhailArkhipov: Originally that's exactly what I did in managed, then discovered that LogLevel is already used for different things in host and also .NET Core uses the term for severity level. So I had to come up with something different. Oh well.

```
namespace Microsoft.Extensions.Logging {
    //
    // Summary:
    //     Defines logging severity levels.
    public enum LogLevel

```

PR #91, opened 2016-09-30
Session management and reconnection (R-Host)

int19h:
Add the ability to load R state from Rdata on startup, and save it on graceful shutdown.

Add a protocol message to request shutdown with or without saving state.

Refactor host termination message: now only used on graceful shutdown to indicate whether state was saved or not.

Add the ability to specify idle timeout, and automatically shutdown the process and save state when it's triggered.

Fix race condition between evaluating R code, and R loading the standard library on startup.

Fix logging verbosity not parsed from command line arguments.

### DIFF
--- a/src/host.cpp
+++ b/src/host.cpp
@@ -43,6 +43,16 @@ namespace rhost {
         boost::signals2::signal<void()> callback_started;
         boost::signals2::signal<void()> readconsole_done;
 
+        fs::path rdata;
+        std::atomic<bool> shutdown_requested = false;
+
+        bool is_r_ready = false;
+        std::mutex is_r_ready_lock;
+        std::condition_variable is_r_ready_cond;
+
+        std::mutex idle_timer_lock;
+        std::chrono::steady_clock::time_point idling_since;
+
         DWORD main_thread_id;
         std::atomic<bool> is_waiting_for_wm = false;
         bool allow_callbacks = true, allow_intr_in_CallBack = true;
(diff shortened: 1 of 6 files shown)


int19h:
The most important changes here relate to https://github.com/Microsoft/RTVS/issues/2386, but there's also a number of important fixes for bugs that were surfaced as part of the more basic reconnection scenarios. I would like to see all this code in, so that it does not bitrot in a separate branch. However, the timeout and state saving logic is not actually used by the broker currently.

int19h:
Corresponding RTVS PR: https://github.com/Microsoft/RTVS/pull/2413

int19h:
Ping?

src/host.cpp
```diff
@@ -185,6 +190,98 @@ namespace rhost {
...
+            for (;;) {
+                std::chrono::steady_clock::time_point idling_since;
+                {
+                    std::lock_guard<std::mutex> lock(idle_timer_lock);
+                    idling_since = host::idling_since;
+                }
+
+                auto delta = std::chrono::steady_clock::now() - idling_since;
+                if (delta > idle_timeout) {
+                    request_shutdown(true);
```
MikhailArkhipov: What is the default timeout?

int19h: The default is no timeout at all. This is represented by a 0-second timeout, but the background watcher thread isn't even started in this case:

https://github.com/Microsoft/R-Host/pull/91/files#diff-59b9dbd011d28180f916f3a21d1af75fR1020

MikhailArkhipov: We want to make sure session won't go down if user went home and left VS up and running. In the morning they will expect session to be still around. Although I guess there is a heartbeat between processes?

huguesv:
:+1:

PR #95, opened 2016-10-10
Logging

MikhailArkhipov:
Buildable in Dev15 Preview 5 using Dev14


src/Microsoft.R.Host.vcxproj
```diff
@@ -110,14 +107,12 @@
...
       <AdditionalIncludeDirectories>..\lib\picojson;..\lib\websocketpp</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
-      <DisableSpecificWarnings>4200</DisableSpecificWarnings>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <AdditionalLibraryDirectories>..\lib</AdditionalLibraryDirectories>
-      <AdditionalDependencies>R.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;Rstrtmgr.lib;%(AdditionalDependencies)</AdditionalDependencies>
-      <StackReserveSize>20971520</StackReserveSize>
```
int19h: If I remember correctly, `StackReserveSize` was necessary for R to function properly with deeply nested call stacks.

src/Microsoft.R.Host.vcxproj
```diff
@@ -110,14 +107,12 @@
       <SDLCheck>false</SDLCheck>
       <AdditionalIncludeDirectories>..\lib\picojson;..\lib\websocketpp</AdditionalIncludeDirectories>
```
int19h: While you're at it, can you remove `websocketpp` from include path?

src/Microsoft.R.Host.vcxproj
```diff
@@ -130,7 +125,6 @@
...
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
-      <DisableSpecificWarnings>4200</DisableSpecificWarnings>
```
int19h: This is suppressing a warning for `class message` in `message.h` - probably still necessary.

MikhailArkhipov: it was complaining about char data[] - can it be changed to char* ?

src/Microsoft.R.Host.vcxproj
```diff
@@ -182,15 +175,14 @@
...
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
-      <DisableSpecificWarnings>4200</DisableSpecificWarnings>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
       <GenerateDebugInformation>true</GenerateDebugInformation>
       <EnableCOMDATFolding>true</EnableCOMDATFolding>
       <OptimizeReferences>true</OptimizeReferences>
       <AdditionalLibraryDirectories>..\lib</AdditionalLibraryDirectories>
-      <AdditionalDependencies>R.lib;Rgraphapp.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;dbghelp.lib;Rstrtmgr.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
int19h: It seems that we have `Rstrtmgr.lib` in list of libraries for some configurations, but not others - looks strange. I would expect it to either be needed everywhere, or be unused everywhere. What is it for?

MikhailArkhipov: I added it to Debug/Release x64, It doesn't seem we've been maintaining Win32 section. I'll double check. Opening project in Dev15 messed up some things, I had to manually restore.

src/message.h
```diff
@@ -30,7 +30,7 @@ namespace rhost {
 
         struct message_repr {
             boost::endian::little_uint64_buf_t id, request_id;
-            char data[];
```
MikhailArkhipov: OK. I remember the meaning is subtly different. I'll revert.

src/Microsoft.R.Host.vcxproj
int19h: Ah. I guess we could just remove these configurations then. Unless we plan to ever build a 32-bit host?

src/message.h
MikhailArkhipov: Yes, you are right, it does matter in structures

int19h: (The reason for the warning is that this is technically a non-standard C++ extension. However, all C++ compilers support it in practice.)

int19h:
👍

PR #101, opened 2016-11-11
Ensure the png device is closed when the ide device is closed.

huguesv:
Fixes https://github.com/Microsoft/RTVS/issues/2578


int19h:
👍

PR #102, opened 2016-11-16
Adds suppress ui flag to command line.

karthiknadig:
https://github.com/Microsoft/RTVS/issues/2702

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -46,6 +46,7 @@ namespace rhost {
         log::log_verbosity log_level;
         std::chrono::seconds idle_timeout;
         std::vector<std::string> unrecognized;
+        bool suppress_ui;
         int argc;
         std::vector<char*> argv;
     };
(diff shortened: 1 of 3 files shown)


src/main.cpp
```diff
@@ -67,10 +68,12 @@ namespace rhost {
             idle_timeout("rhost-idle-timeout", po::value<std::chrono::seconds::rep>(), (
                 "Shut down the host if it is idle for the specified duration in seconds. "
                 "If " + rdata.long_name() + " was specified, save workspace before exiting."
-                ).c_str());
+                ).c_str()),
+            suppress_ui("suppress-ui", new po::untyped_value(true),
```
int19h: I believe you can just omit the value here, and only have the name and the docstring.

src/main.cpp
```diff
@@ -122,6 +125,12 @@ namespace rhost {
             args.idle_timeout = std::chrono::seconds(n);
         }
 
+        if (vm.count(suppress_ui.long_name())) {
```
int19h: `args.suppress_ui = vm.count(...) != 0`

src/main.cpp
karthiknadig: Does't look like it supports that form of constructor:
```
option_description();
option_description(const char* name, const value_semantic* s);
option_description(const char* name, const value_semantic* s, const char* description);
```

int19h:
👍

PR #104, opened 2016-12-19
rtvs::fetch_file progress

MikhailArkhipov:


src/stdafx.h
```diff
@@ -63,6 +63,7 @@
 #include "boost/uuid/uuid_io.hpp"
 #include "boost/uuid/uuid_generators.hpp"
 
+#define PICOJSON_USE_INT64
```
int19h: This is no longer standard JSON, and I'd prefer to keep it standard. Use double instead.

MikhailArkhipov: 64 bit ints are non-standard... What century is it ;-)

int19h:
👍

PR #106, opened 2017-01-26
Fixes 32bit build

karthiknadig:
For Sprint 11:
* Adds x86 library for R and Rgraphapp.
* Fixes the 32bit build of RHost.


int19h:
👍

PR #107, opened 2017-01-31
Notify when there is a UI pop-up.

karthiknadig:
R-Host side changes to address this : https://github.com/Microsoft/RTVS/issues/3069

### DIFF
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -74,10 +74,19 @@ namespace rhost {
             size_t len = 0;
             if (hWnd != nullptr && hWndParent == nullptr && 
                 hWndTitleMap.find(hWnd) == hWndTitleMap.end() && 
-                SUCCEEDED(StringCchLengthA(lpWindowName, STRSAFE_MAX_CCH, &len)) && len > 0) {
+                SUCCEEDED(StringCbLengthA(lpWindowName, STRSAFE_MAX_CCH, &len)) && len > 0) {
 
-                std::string windowTitle = std::string(lpWindowName);
-                std::wstring_convert<std::codecvt_utf8<char>, char> strConverter;
+                size_t size = MultiByteToWideChar(CP_ACP, 0, lpWindowName, len, NULL, 0);
+                if (size == 0) {
+                    return hWnd;
+                }
+
+                std::unique_ptr<wchar_t[]> wstr(new wchar_t[size + 1]);
+                MultiByteToWideChar(CP_ACP, 0, lpWindowName, len, wstr.get(), size);
+                wstr[size] = '\0';
+
+                std::wstring windowTitle = std::wstring(wstr.get());
+                std::wstring_convert<std::codecvt_utf8<wchar_t>, wchar_t> strConverter;
 
                 hWndTitleMap[hWnd] = strConverter.to_bytes(windowTitle);
                 std::string text = "Created window - " + hWndTitleMap[hWnd];
--- a/src/detours.cpp
+++ b/src/detours.cpp
@@ -52,7 +52,7 @@ namespace rhost {
         }
 
         // MessageBoxW
-        decltype(MessageBoxW) *pMessageBoxW = NULL;
+        decltype(MessageBoxW) *pMessageBoxW = nullptr;
         int WINAPI DetourMessageBoxW(HWND hWnd, LPCWSTR lpText, LPCWSTR lpCaption, UINT uType) {
             char ch[1000];
             wcstombs(ch, lpText, _countof(ch));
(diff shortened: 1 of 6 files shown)


src/detours.h
```diff
@@ -23,7 +23,7 @@
 #pragma once
 namespace rhost {
     namespace detours {
-        void init_ui_detours();
+        void init_ui_detours(bool isRemote);
```
int19h: `is_remote`

src/detours.cpp
```diff
@@ -52,31 +52,89 @@ namespace rhost {
...
+        decltype(MessageBoxW) *pMessageBoxW = nullptr;
         int WINAPI DetourMessageBoxW(HWND hWnd, LPCWSTR lpText, LPCWSTR lpCaption, UINT uType) {
             char ch[1000];
             wcstombs(ch, lpText, _countof(ch));
             return HostMessageBox(ch, uType);
         }
 
         // MessageBoxA
-        decltype(MessageBoxW) *pMessageBoxA = NULL;
+        decltype(MessageBoxA) *pMessageBoxA = nullptr;
```
int19h: Do we actually need to detour this? I thought that all the A functions will call W functions under the hood anyway.

src/detours.cpp
```diff
@@ -52,31 +52,89 @@ namespace rhost {
...
+        HWND WINAPI DetourCreateWindowExA(DWORD dwExStyle, LPCSTR lpClassName, LPCSTR lpWindowName, DWORD dwStyle, int X, int Y, int nWidth, int nHeight, HWND hWndParent, HMENU hMenu, HINSTANCE hInstance, LPVOID lpParam) {
+            HWND hWnd = pCreateWindowExA(dwExStyle, lpClassName, lpWindowName, dwStyle, X, Y, nWidth, nHeight, hWndParent, hMenu, hInstance, lpParam);
+
+            // A window without parent or a window created to show message
+            size_t len = 0;
+            if (hWnd != nullptr && hWndParent == nullptr && 
+                hWndTitleMap.find(hWnd) == hWndTitleMap.end() && 
+                SUCCEEDED(StringCchLengthA(lpWindowName, STRSAFE_MAX_CCH, &len)) && len > 0) {
+
+                hWndTitleMap[hWnd] = std::string(lpWindowName);
```
int19h: This isn't UTF-8, and so will produce unparseable JSON for non-ASCII stuff.

src/detours.cpp
```diff
@@ -52,31 +52,89 @@ namespace rhost {
...
+                rhost::host::send_notification("!!", text);
+            }
+
+            return hWnd;
+        }
+
+        decltype(DestroyWindow) *pDestroyWindow = nullptr;
+        BOOL WINAPI DetourDestroyWindow(HWND hWnd) {
+            BOOL stat = pDestroyWindow(hWnd);
+            if (hWndTitleMap.find(hWnd) != hWndTitleMap.end()) {
```
int19h: The usual pattern is to save the iterator returned by `find`, so that you don't have to do the second lookup with `[]`, but just use the iterator to access the value via `iter->second`. You can then also use it for `erase`.

src/main.cpp
```diff
@@ -39,6 +39,9 @@ using namespace rhost::log;
 using namespace rhost::util;
 namespace po = boost::program_options;
 
+// ID for the timer 
+#define IDT_RESET_TIMER 1
```
int19h: `const UINT_PTR`.

int19h: (and inside the namespace)

src/detours.cpp
karthiknadig: MBA and MBW eventually call a internal W function. MBA does not internally call MBW.

karthiknadig:
Ping

src/detours.cpp
```diff
@@ -52,31 +52,94 @@ namespace rhost {
...
+            HWND hWnd = pCreateWindowExA(dwExStyle, lpClassName, lpWindowName, dwStyle, X, Y, nWidth, nHeight, hWndParent, hMenu, hInstance, lpParam);
+
+            // A window without parent or a window created to show message
+            size_t len = 0;
+            if (hWnd != nullptr && hWndParent == nullptr && 
+                hWndTitleMap.find(hWnd) == hWndTitleMap.end() && 
+                SUCCEEDED(StringCchLengthA(lpWindowName, STRSAFE_MAX_CCH, &len)) && len > 0) {
+
+                std::string windowTitle = std::string(lpWindowName);
+                std::wstring_convert<std::codecvt_utf8<char>, char> strConverter;
```
MikhailArkhipov: It if is 'A' then the encoding will be `CP_ACP` (ie console or 'non-Unicode app` code page and not `UTF-8`. Windows doesn't support UTF8 in console.

karthiknadig: what is the correct convert for this?

src/detours.cpp
```diff
@@ -76,7 +76,10 @@ namespace rhost {
                 hWndTitleMap.find(hWnd) == hWndTitleMap.end() && 
                 SUCCEEDED(StringCchLengthA(lpWindowName, STRSAFE_MAX_CCH, &len)) && len > 0) {
 
-                hWndTitleMap[hWnd] = std::string(lpWindowName);
+                std::string windowTitle = std::string(lpWindowName);
+                std::wstring_convert<std::codecvt_utf8<char>, char> strConverter;
```
int19h: This converts from UTF-8 to wchar_t. But the input in this case is not UTF-8.

Since this code is heavily Win32-dependent in any case, and input comes from Win32 also, I think it's easier to just use Win32 functions for this.

karthiknadig: What is the correct way to convert this? I am not familiar with these conversions.

int19h: https://msdn.microsoft.com/en-us/library/windows/desktop/dd319072(v=vs.85).aspx

karthiknadig:
I have made the changes to add --repl, string conversions. If there are no more comments i will merge.

int19h:
There's a comment on the other review (in RTVS) regarding the switch name and meaning - can you take a look?

int19h:
https://github.com/Microsoft/RTVS/pull/3087#commitcomment-20702960

karthiknadig:
updated the name.

int19h:
👍

PR #111, opened 2017-02-21
ifdef windows specific code

karthiknadig:


src/Microsoft.R.Host.vcxproj
```diff
@@ -103,7 +103,7 @@
       </PrecompiledHeader>
       <WarningLevel>Level3</WarningLevel>
       <Optimization>Disabled</Optimization>
-      <PreprocessorDefinitions>WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
+      <PreprocessorDefinitions>TRACE_JSON;WIN32;_DEBUG;_CONSOLE;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```
int19h: `WIN32` is defined without the underscore here, but you use `_WIN32` below - is that intentional?

src/transport.cpp
```diff
@@ -98,9 +98,9 @@ namespace rhost {
             setmode(fileno(stdout), _O_BINARY);
 
             // Duplicate and stash away handles for original stdin & stdout.
-            input = fdopen(dup(fileno(stdin)), "rb");
+            input = fdopen(_dup(fileno(stdin)), "rb");
```
int19h: It needs to be `dup` to be cross-plat (it's a POSIX function).

src/Microsoft.R.Host.vcxproj
karthiknadig: VS C++ console project template has that by default for 32 bit builds.

src/transport.cpp
karthiknadig: I see, I will change that.

int19h:
👍

MikhailArkhipov:
Master is RC2 at the moment, shall we take this or delay?

int19h:
It's being merged into a separate branch, not master.

MikhailArkhipov:
OK

PR #112, opened 2017-02-23
Fix #3209: After running tests R.Host.exe processes do not terminate

int19h:
Fix race condition in shutdown while waiting for a response to a request.

Do not allow blocking requests (that wait for responses) after the client has disconnected.


int19h:
Also see https://github.com/Microsoft/RTVS/pull/3228

MikhailArkhipov:
👍

karthiknadig:
👍

PR #114, opened 2017-03-02
Building with gcc.

karthiknadig:
These were the changes needed to get RHost building with **gcc-6.1** and **boost-1.63**. Some of the features needed were disabled. There are tasks to find and add linux equivalents for those.
https://github.com/Microsoft/RTVS/issues/3247
https://github.com/Microsoft/RTVS/issues/3103


src/grdeviceside.cpp
```diff
@@ -468,8 +468,11 @@ namespace rhost {
             ///////////////////////////////////////////////////////////////////////
 
             std::unique_ptr<ide_device> ide_device::create(const boost::uuids::uuid& device_id, std::string device_type, double width, double height, double resolution) {
+#ifdef _WIN32
                 pDevDesc dd = static_cast<pDevDesc>(rhost::msvcrt::calloc(1, sizeof(DevDesc)));
-
+#else
+                pDevDesc dd = static_cast<pDevDesc>(calloc(1, sizeof(DevDesc)));
```
int19h: Since this is used several times, and the two versions are really the same function (just from different DLLs), can you make a wrapper that contains the ifdef inside? Something like `R_calloc`.

src/host.cpp
```diff
@@ -1011,7 +1011,11 @@ namespace rhost {
                         continue;
                     }
 
+#ifdef _WIN32
                     strcpy_s(buf, len, s.c_str());
+#else
+                    strcpy(buf, s.c_str());
```
int19h: I would suggest that we make our own implementation of `strcpy_s` here. For now it can be just a simple wrapper over `strcpy` that ignores the buffer length parameter, but later we would add proper checking to it. Ditto for other `_s` functions.

src/log.cpp
```diff
@@ -139,8 +142,12 @@ namespace rhost {
                 stackdump_filename = log_dir / (filename + ".stack.dmp");
                 fulldump_filename = log_dir / (filename + ".full.dmp");
             }
-        
+
+#ifdef _WIN32
             logfile = _fsopen(log_filename.make_preferred().string().c_str(), "wc", _SH_DENYWR);
+#else
+            logfile = fopen(log_filename.make_preferred().string().c_str(), "wc");
```
int19h: "c" is a VC-specific open mode (it disables buffering). It won't be supported on Linux. So this should be just "w".

karthiknadig: http://man7.org/linux/man-pages/man3/fopen.3.html I see mode extensions at the bottom. Are those not supported?

src/grdeviceside.cpp
karthiknadig: I have done this for a few others. I will make the change.

src/stdafx.h
```diff
@@ -81,29 +82,62 @@
...
+#define RHOST_NOTHROW
 #pragma warning Unknown DLL import/export.
 #endif
 
 #ifdef _WIN32
 #define RHOST_MAX_PATH MAX_PATH
+typedef DWORD threadid_t;
+#define RhostGetCurrentThreadId GetCurrentThreadId
+#define RHOST_TRY __try
+#define RHOST_FINALLY_START __finally {
```
int19h: Can you remove the `{` from the macro, and also define it for non-Win32 as simply `{`? Then the braces would remain as they were at the place where it's used. And we won't need `RHOST_FINALLY_END` at all, it'll just be `}`. Since Linux doesn't have anything like SEH, these definitions should actually be correct, and not just temp stubs.

Also, can you rename this to `SEH_TRY` and `SEH_FINALLY`, to make it sure that's not for normal C++ exceptions?

src/stdafx.h
```diff
@@ -81,29 +82,62 @@
 
 #include "zip.h"
 
+namespace fs = std::tr2::sys;
```
int19h: Let's just use Boost.FS everywhere. No particular reason to try to jiggle two subtly different implementations.

src/stdafx.h
```diff
@@ -81,29 +82,62 @@
...
 #else
+#include "boost/filesystem.hpp"
+#include <stdarg.h>
```
int19h: This one is actually a standard header, so it should go into the unconditional block at the beginning of the file. Also, can you make it `<cstdarg>`, for consistency with the rest?

src/stdafx.h
```diff
@@ -81,29 +82,62 @@
...
+#include <unistd.h>
+namespace fs = boost::filesystem;
 
-//#include <unistd.h>
+#include <sys/syscall.h>
```
int19h: This is Linux-specific. We'll probably want OS X and others, too, so let's use something more portable.

Actually, looking at where it's used, on Win32 it's only needed to post a message into the message queue of the thread. It is likely that any equivalent code on Linux and elsewhere will not use a thread ID for that purpose. So for now, I would suggest removing this, and just ifdef'ing all uses of `main_thread_id` on non-Win32. We'll figure out the proper abstraction for this later.

src/stdafx.h
```diff
@@ -81,29 +82,62 @@
...
 #endif
 
-namespace fs = std::tr2::sys;
+#endif
 
 #if defined(_MSC_VER)
 #define RHOST_EXPORT __declspec(dllexport)
 #define RHOST_IMPORT __declspec(dllimport)
+#define RHOST_NORETURN __declspec(noreturn)
+#define RHOST_NOTHROW __declspec(nothrow)
```
int19h: I think we can just start using `noexcept` instead. All the compilers we care about should support it. The `SCOPE_WARDEN` code was from older times.

src/log.cpp
int19h: It does list "c", but it appears to be something completely different from VC.

int19h:
👍

PR #115, opened 2017-03-03
zip feature for rhost linux.

karthiknadig:
Enables zip feature for RHost linux. Requires the following libraries
**libzip-dev**: library for reading, creating, and modifying zip archives (development)
**libzip4**: library for reading, creating, and modifying zip archives (runtime)


int19h:
👍

PR #117, opened 2017-03-10
Some more minor fixes to build on linux.

karthiknadig:


src/r_util.cpp
```diff
@@ -516,8 +516,9 @@ namespace rhost {
             std::vector<std::wstring> files;
             r_top_level_exec([&]() {
                 for (R_len_t i = 0; i < len; ++i) {
-                    const wchar_t* file_path = Rf_wtransChar(STRING_ELT(paths, i));
-                    files.emplace_back(file_path);
+                    const char* file_path_str = Rf_translateCharUTF8(STRING_ELT(paths, i));
+                    fs::path file_path(file_path_str);
```
int19h: Does this work correctly on Windows? My understanding was that when Boost.Filesystem gets a char* argument, it assumes it is in whatever the appropriate encoding for filesystem is in the system locale (i.e. CP_ACP for Win32, probably), rather than UTF-8.

So we might have to ifdef our way around this, using wchar_t on Win32 and char on Linux (and assume UTF-8 locale on the latter, which is usually true).

karthiknadig: you are right. I missed the conversion. I will create a util function to do this correctly.

---
In reply to: [105302683](https://github.com/Microsoft/R-Host/pull/117#discussion_r105302683) [](ancestors = 105302683)

int19h: Don't go through the locale encoding though, because that would be potentially lossy (your codepage might not be able to represent an arbitrary Unicode pathname, but your filesystem might have files with such names).

int19h:
👍

PR #119, opened 2017-03-20
Support building with MSYS2 MinGW on Win32

int19h:
Merging the work so far for the sake of convenience. Still work in progress: needs to be wired up to MSBuild, so that normal solution build will actually use MSYS.


.gitignore
```diff
@@ -1,71 +1,17 @@
...
-*.dotCover
-
-# NuGet Packages Directory
-NugetPackages/*
-
-# Windows Azure Build Output
-csx/
-*.build.csdef
-
-# Others
```
AlexanderSher: I like this part :)

int19h: Most of it was unnecessary for our existing builds. It's just some template that tries to include as many VS artifacts as possible, most of which aren't applicable to C++ projects in general.

CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
+cmake_minimum_required(VERSION 2.7)
+project(R-Host)
+
+if(${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_BINARY_DIR})
+    message(FATAL_ERROR "In-source builds are not allowed. Please invoke CMake from ./build, or use ./build.sh helper script.")
+endif()
+
+if(CMAKE_BUILD_TYPE MATCHES DEBUG)
+    set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_HOME_DIRECTORY}/bin/Release)
```
AlexanderSher: Where `CMAKE_HOME_DIRECTORY` is defined?

int19h: CMake itself defines it - it's where CMakeLists.txt lives.

AlexanderSher: Ok, so how we're going to build it from RTVS?

int19h: Our .vcxproj is going to invoke CMake and make from MSYS2. I'm still working on that, and it's not a part of this PR. This one just lets you build it manually from MSYS2 shell.

(Note that this is a PR into another branch - the one Karthik created for cross-plat work - not into master.)

AlexanderSher:
👍

karthiknadig:
:shipit:

src/detours.cpp
```diff
@@ -138,14 +138,14 @@ namespace rhost {
 #ifdef _WIN32
             MH_Initialize();
 
-            MH_CreateHook(&MessageBoxW, &DetourMessageBoxW, reinterpret_cast<LPVOID*>(&pMessageBoxW));
-            MH_CreateHook(&MessageBoxA, &DetourMessageBoxA, reinterpret_cast<LPVOID*>(&pMessageBoxA));
+            MH_CreateHook((void*)&MessageBoxW, (void*)&DetourMessageBoxW, reinterpret_cast<LPVOID*>(&pMessageBoxW));
```
karthiknadig: >(void*) [](start = 47, length = 7)

static_cast?

int19h: static_cast won't work - data and function pointers are not convertible in standard C. I think reinterpret_cast might work, let me give it a try.

PR #120, opened 2017-03-22
Removing suggest_mro from RHost

karthiknadig:
Moved this check to client.
https://github.com/Microsoft/RTVS/issues/3319


int19h:
👍

PR #122, opened 2017-03-27
Linux host

karthiknadig:


src/host.cpp
```diff
@@ -1084,18 +1109,13 @@ namespace rhost {
...
-            rp.YesNoCancel = YesNoCancel;
-            rp.Busy = Busy;
-
+            setCallbacksWindows(rp);
             char* dllVersion = getDLLVersion();
+            send_notification("!Microsoft.R.Host", 1.0, dllVersion);
 #else
-            char* dllVersion = "";
+            setCallbacksPOSIX();
+            send_notification("!Microsoft.R.Host", 1.0, "3.2.3");
```
MikhailArkhipov: Sure about 3.2.3?

src/main.cpp
```diff
@@ -189,25 +195,22 @@ namespace rhost {
             return;
         }
 
-#ifdef _WIN32
         static const char mro_banner[] = "Check out Microsoft's enhanced R distribution at https://aka.ms/mrclient. \n\n";
```
MikhailArkhipov: This technically should be localizable and hence should be in the client. Host can send notification to the client to drop banner to REPL.

src/main.cpp
```diff
@@ -258,25 +255,82 @@ namespace rhost {
...
+
+            bool ok = r_top_level_exec([&] {
+                R_RestoreGlobalEnvFromFile(s.c_str(), R_FALSE);
+            });
+
+            log::logf(log_verbosity::minimal, ok ? "Workspace loaded successfully.\n" : "Failed to load workspace.\n");
+        }
+
+        Rf_mainloop();
+
```
MikhailArkhipov: MRO does exist on Linux. We should still be suggesting it. https://mran.microsoft.com/download/

src/util.h
```diff
@@ -118,17 +121,31 @@ namespace rhost {
...
+#ifdef _WIN32
         std::string Rchar_to_utf8(const char* buf, size_t len);
 
         inline std::string Rchar_to_utf8(const std::string& s) {
             return Rchar_to_utf8(s.data(), s.size());
         }
+        std::string from_utf8(const std::string& u8s);
+#else
+        inline std::string Rchar_to_utf8(const char* buf, size_t len) {
+            return std::string(buf, len);
```
int19h: On any modern (like last several years) Linux distro it is. f it's not, we're in a world of hurt anyway because e.g. pathnames are not Unicode either.

src/host.cpp
```diff
@@ -921,7 +921,11 @@ namespace rhost {
             do_r_callback(true);
         }
 
-        extern "C" int R_ReadConsole(const char* prompt, char* buf, int len, int addToHistory) {
+#ifdef _WIN32
+        extern "C" int R_ReadConsole(const char* prompt, char * buf, int len, int addToHistory) {
```
int19h: Use a typedef for `buf` instead, and do ifdef around that - it'll be a bit clearer.

src/host.cpp
```diff
@@ -1075,6 +1079,27 @@ namespace rhost {
             }
         }
 
+        void setCallbacksWindows(structRstart& rp) {
```
int19h: This feels like the ifdef should be around the entire function (including the definition), since they don't do anything on other platforms.

src/host.cpp
int19h: We'll need to figure out how to get the version right. But that can be a separate change.

src/main.cpp
```diff
@@ -25,7 +25,11 @@
 #include "log.h"
 #include "r_util.h"
 #include "host.h"
+
+#ifdef _WIN32
```
int19h: I'm thinking maybe instead of repeating this everywhere, put the guards inside `Rapi.h` itself, around its entire content?

src/host.cpp
karthiknadig: You are correct. For now this is a TODO. I will add a comment there.

src/main.cpp
karthiknadig: In main this has been moved to client. So it is localize-able.

src/main.cpp
karthiknadig: We do suggest it, but we do this from the client thing after connecting to host. In linux eval cannot be called before `Rf_mainloop`.

src/main.cpp
karthiknadig: You are correct. Will do.

karthiknadig:
Addresses this issue: https://github.com/Microsoft/RTVS/issues/3358

src/host.cpp
```diff
@@ -38,6 +38,12 @@ using namespace rhost::json;
 using namespace rhost::blobs;
 using namespace rhost::protocol;
 
+#ifdef _WIN32
+typedef char rhost_buf;
```
int19h: This should probably be something like `ReadConsole_buf_t`

src/host.cpp
```diff
@@ -1079,26 +1082,24 @@ namespace rhost {
             }
         }
 
-        void setCallbacksWindows(structRstart& rp) {
 #ifdef _WIN32
+        void setCallbacksWindows(structRstart& rp) {
```
int19h: Sorry, didn't notice it on the last pass, but can you change the identifier to conform to the `underscore_lowercase` convention that we use elsewhere, for consistency?

int19h:
A few minor comments, otherwise 👍

PR #124, opened 2017-03-30
Switch to MSYS2 / MinGW

int19h:
Putting this in a separate branch from `crossplat` for now, because with these changes, you *must* have MSYS2 to build (.vcxproj is now wired to invoke CMake).


karthiknadig:
👍

PR #127, opened 2017-04-04
Assorted MSYS and Linux build fixes and improvements

int19h:
With these changes, you should be able to build on Linux using `build.sh`.


int19h:
Note: I renamed `internal_rapi.h` to `r_api.h`, because it's not just exclusively for internal APIs now - it also brings in the public R headers.

karthiknadig:
👍

PR #131, opened 2017-04-05
Loading R library instead of linking it

karthiknadig:
Initial changes for review by @int19h 
https://github.com/Microsoft/RTVS/issues/3386
https://github.com/Microsoft/RTVS/issues/3395

image


src/loadr.cpp
```diff
@@ -0,0 +1,83 @@
...
+*
+* You should have received a copy of the GNU General Public License
+* along with Microsoft R Host.  If not, see <http://www.gnu.org/licenses/>.
+*
+* ***************************************************************************/
+
+#define RHOST_NO_API_REDIRECT
+#include "stdafx.h"
+
+HMODULE r_module = nullptr;
```
int19h: These should also go into the namespace.

src/loadr.cpp
```diff
@@ -0,0 +1,83 @@
...
+HMODULE rgraphapp_module = nullptr;
+
+#define _GET_PROC(m, api) _RAPI_PTR(api) = get_proc<decltype(api)>(m, _RAPI_STR(api));
```
int19h: Identifiers starting with underscore followed by capital letter are reserved to the compiler and the standard library.

In this case, since these macros are contained within the .cpp file rather than a header, there's no reason to obfuscate their names.

src/loadr.cpp
```diff
@@ -0,0 +1,83 @@
...
+    namespace rapi {
+        _RAPI_SET(_RAPI_DEFINE_NULLPTR);
+        _RGRAPHAPPAPI_SET(_RAPI_DEFINE_NULLPTR);
+
+        namespace {
+            template<typename T>
+            T* get_proc(HMODULE module, std::string proc) {
+                T* ptr = nullptr;
+#ifdef _WIN32
+                ptr = (T*)GetProcAddress(module, proc.c_str());
```
int19h: `reinterpret_cast`

src/loadr.cpp
```diff
@@ -0,0 +1,83 @@
...
+#endif
+
+namespace rhost {
+    namespace rapi {
+        _RAPI_SET(_RAPI_DEFINE_NULLPTR);
+        _RGRAPHAPPAPI_SET(_RAPI_DEFINE_NULLPTR);
+
+        namespace {
+            template<typename T>
+            T* get_proc(HMODULE module, std::string proc) {
```
int19h: This should be `const std::string&`.

Although it looks like we're always actually passing string literals to in? In which case it might as well be `const char*`.

src/loadr.cpp
```diff
@@ -0,0 +1,83 @@
...
+                _RGRAPHAPPAPI_SET(_RGRAPHAPPAPI_GET_PROC);
+            }
+
+            void internal_unload_r_apis() {
+                _RAPI_SET(_RAPI_UNLOAD);
+            }
+        }
+        
+
+        void load_r_apis(fs::path r_dll_dir) {
```
int19h: `const fs::path&`

src/loadr.h
```diff
@@ -0,0 +1,278 @@
...
+*
+* You should have received a copy of the GNU General Public License
+* along with Microsoft R Host.  If not, see <http://www.gnu.org/licenses/>.
+*
+* ***************************************************************************/
+
+#pragma once
+#include "stdafx.h"
+
+#define _RAPI_PTR(api) fp_##api
```
int19h: Same thing about the `_` prefix. Let's do something like `RAPI_IMPL_...`

src/loadr.h
```diff
@@ -0,0 +1,278 @@
...
+        // Rgraphapp.dll/Rgraphapp.so APIs
+        _RGRAPHAPPAPI_SET(_RAPI_DECL_END);
+#endif
+        void load_r_apis(fs::path r_dll_dir);
+        void unload_r_apis();
+    }
+}
+
+#ifndef RHOST_NO_API_REDIRECT
+#define CAR rhost::rapi::_RAPI_PTR(CAR)
```
int19h: Do these have to be macros? It feels like if you drop the `fp_` prefix in `_RAPI_PTR`, then we'll just have variables with the right names directly in the namespace. And because those variables will have function pointer types, they will be callable directly. You'll only need to do `using namespace rhost::rapi` in other files.

karthiknadig: I though that would lead to ambiguous code:
```
#include "r.h"
#include "loadr.h"
using rhost::rapi;

// which CAR is being called here
CAR();
```

int19h: It should work if `using namespace` is placed inside `namespace` - i.e.:
```c++
namespace rhost {
    namespace host {
        using namespace rapi;
        ...
```

karthiknadig: Ran into a few complications with this. One thing was the get proc. Using decltype was ambiguous, compiler got confused with which deceleration to use( the actual function or the function ptr). Additionally, we would have to update all usages of object/data member exports to (*datamember). So we will need the defines at least for the data members, to keep it clean.

PR #133, opened 2017-04-26
Support for R_33_GE_VERSION and R_34_GE_VERSION

karthiknadig:
Helps with addressing this issue: https://github.com/Microsoft/RTVS/issues/3441

### DIFF
--- a/src/Microsoft.R.Host.vcxproj
+++ b/src/Microsoft.R.Host.vcxproj
@@ -108,7 +108,6 @@
       <AdditionalIncludeDirectories>..\lib\picojson</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
       <DisableSpecificWarnings>4200</DisableSpecificWarnings>
-      <PreprocessToFile>false</PreprocessToFile>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
@@ -134,7 +133,6 @@
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
       <DisableSpecificWarnings>4200</DisableSpecificWarnings>
       <ControlFlowGuard>false</ControlFlowGuard>
-      <PreprocessToFile>false</PreprocessToFile>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
@@ -165,7 +163,6 @@
       <AdditionalIncludeDirectories>..\lib\picojson</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
       <DisableSpecificWarnings>4200</DisableSpecificWarnings>
-      <PreprocessToFile>false</PreprocessToFile>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
@@ -196,7 +193,6 @@
       <RuntimeLibrary>MultiThreaded</RuntimeLibrary>
       <DisableSpecificWarnings>4200</DisableSpecificWarnings>
       <ControlFlowGuard>Guard</ControlFlowGuard>
-      <PreprocessToFile>false</PreprocessToFile>
     </ClCompile>
     <Link>
       <SubSystem>Console</SubSystem>
@@ -218,7 +214,7 @@
   </ItemDefinitionGroup>
   <ItemGroup>
     <ClCompile Include="blobs.cpp" />
-    <ClCompile Include="devicewrapper.cpp" />
+    <ClCompile Include="devdesc_wrapper.cpp" />
     <ClCompile Include="exports.cpp" />
     <ClCompile Include="grdeviceside.cpp" />
     <ClCompile Include="grdevicesxaml.cpp" />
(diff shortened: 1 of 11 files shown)


src/Microsoft.R.Host.vcxproj
```diff
@@ -108,6 +108,7 @@
       <AdditionalIncludeDirectories>..\lib\picojson</AdditionalIncludeDirectories>
       <RuntimeLibrary>MultiThreadedDebug</RuntimeLibrary>
       <DisableSpecificWarnings>4200</DisableSpecificWarnings>
+      <PreprocessToFile>false</PreprocessToFile>
```
karthiknadig: I was using this to check the PreProcessor output. Will remove.

src/devicewrapper.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include "msvcrt.h"
+
+namespace rhost {
+    namespace grdevices {
+        devicewrapper::devicewrapper() {
+            _ver = R_GE_getVersion();
+        }
+
+
+        devicewrapper::devicewrapper(pDevDesc dd) {
```
MikhailArkhipov: This is a factory so I'd call it `graphical_device_factory`

huguesv: While there are factory methods on this class, there are a lot more wrapper methods.  They're all expanded using the macros at lines 62-64.

---
In reply to: [113523598](https://github.com/Microsoft/R-Host/pull/133#discussion_r113523598) [](ancestors = 113523598)

huguesv: Since there's a few device related structs/classes, and this is specifically for the DevDesc structure, I would probably name it devdesc_wrapper.

---
In reply to: [113535789](https://github.com/Microsoft/R-Host/pull/133#discussion_r113535789) [](ancestors = 113535789,113523598)

src/devicewrapper.cpp
```diff
@@ -0,0 +1,66 @@
...
+            _ver = R_GE_getVersion();
+            switch (_ver) {
```
huguesv: >switch [](start = 12, length = 6)

Assert in default case and always initialize both pointers (set the unused one to nullptr). If we needed to save memory, we could use a union since only one of dd_11 and dd_12 will ever be used, but union is more bug prone and this is not a memory intensive scenario.

huguesv:
> #define R_GE_version 10

If this isn't used, it's probably best to delete it.

---
Refers to: src/Rgraphicsapi.h:42 in d714ccf. [](commit_id = d714ccf2cad1446c7f123faab9e553129be02d7e, deletion_comment = False)

src/devicewrapper.h
```diff
@@ -0,0 +1,249 @@
...
+    case ver: return RHOST_DEV_DEREF_ARRAY(ver, member); \
+
+#define RHOST_DEV_SET_CASE(ver, member) \
+    case ver: RHOST_DEV_DEREF(ver, member) = value; break; \
+
+#define RHOST_DEV_SET_CASE_ARRAY(ver, member) \
+    case ver: RHOST_DEV_DEREF_ARRAY(ver, member) = value; break; \
+
+#define RHOST_DEV_GET_CASE_COMMON(member) \
+    RHOST_DEV_GET_CASE(R_33_GE_version, member) \
```
huguesv: >R_33_GE_version [](start = 23, length = 15)

since you're using 11 and 12 everywhere else in devicewrapper.h/.cpp, it's probably less confusing if you use 11 and 12 here too, instead of R_33, R_34...

src/devicewrapper.cpp
```diff
@@ -0,0 +1,66 @@
...
+            case 11:
```
huguesv: >case [](start = 12, length = 4)

we support 10 as well.  it's the same struct as 11.  same for the 2 functions below.

src/devicewrapper.h
huguesv: these macros need to include a case statement for version 10 too.  It's probably simpler to add a dd_10 than try to make it reuse dd_11.

---
In reply to: [113544419](https://github.com/Microsoft/R-Host/pull/133#discussion_r113544419) [](ancestors = 113544419)

huguesv:
:clock1:

src/devicewrapper.cpp
huguesv: Maybe union is best after all, if we end up with a field for v10. This will only grow unnecessarily if it's not a union.

---
In reply to: [113539632](https://github.com/Microsoft/R-Host/pull/133#discussion_r113539632) [](ancestors = 113539632)

huguesv:
It will be important to run the IdeGraphicsDeviceTest suite of tests, against R 3.2, 3.3, 3.4.

karthiknadig:
IdeGraphicsDeviceTest Passed for 3.2.5, 3.3.3, and 3.4.0. I tested this by removing all R/MRO/RClient version and only installing the one that i wanted to test.

---
In reply to: [297530457](https://github.com/Microsoft/R-Host/pull/133#issuecomment-297530457) [](ancestors = 297530457)

PR #139, opened 2017-06-04
Linux support, and CMake/MinGW build for Windows

int19h:


src/grdevicesxaml.cpp
```diff
@@ -20,10 +20,10 @@
 *
 * ***************************************************************************/
 
+#if 0
```
int19h: This entire file needs to be reworked to compile with all the recent changes to R API handling. However, in light of https://github.com/Microsoft/RTVS/issues/3596, it doesn't make sense to do it now, since the fix to that bug will require more changes to the same things. Since we're not actually advertising this as a feature, disable it for now until https://github.com/Microsoft/RTVS/issues/3596 is done.

karthiknadig:
R.lib, R.def, etc. files under the `lib` folder. They are not needed since we loadlib R manually now.

PR #140, opened 2017-06-06
Fix for R 3.4 - LoadedDll static member variable was changed from array to a pointer.

karthiknadig:
Using setup_Rmainloop and later calling run_Rmainloop resolves this issue. Both these functions are available in 3.2 and 3.3.


MikhailArkhipov:
👍

PR #141, opened 2017-06-08
Add R bin path to PATH environment variable.

karthiknadig:
https://github.com/Microsoft/RTVS/issues/3592

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -183,6 +183,37 @@ namespace rhost {
 #endif
     }
 
+    void add_dir_to_loader_path(fs::path &r_dir) {
+#ifdef _WIN32
+        wchar_t buff[32767] = {}; // max buffer size for environment variable values on windows.
+        int nchars = GetEnvironmentVariable(L"PATH", buff, ARRAYSIZE(buff));
+        HRESULT err = S_OK;
+        if (nchars == 0) {
+            err = HRESULT_FROM_WIN32(GetLastError());
+            if (err == HRESULT_FROM_WIN32(ERROR_ENVVAR_NOT_FOUND)) {
+                buff[0] = '\0';
+            } else {
+                fatal_error("Failed to get PATH environment variable [GetEnvironmentVariable]: %x\n", err);
+            }
+        }
+
+        std::wstring old_path_env(buff);
+        std::wstring new_path_env = r_dir.make_preferred().wstring() + L";" + old_path_env;
+        if (SetEnvironmentVariableW(L"PATH", new_path_env.c_str()) == 0) {
+            err = HRESULT_FROM_WIN32(GetLastError());
+            fatal_error("Failed to set PATH environment variable [SetEnvironmentVariable]: %x\n", err);
+        }
+#else
+        char *path_env = std::getenv("LD_LIBRARY_PATH");
+        std::string old_path_env = path_env ? std::string(path_env) : std::string();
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -183,6 +183,35 @@ namespace rhost {
...
     }
 
+    void add_dir_to_loader_path(fs::path &r_dir) {
+#ifdef _WIN32
+        wchar_t buff[32767] = {}; // max buffer size for environment variable values on windows.
+        int nchars = GetEnvironmentVariable(L"PATH", buff, ARRAYSIZE(buff));
+        HRESULT err = S_OK;
+        if (nchars == 0) {
+            err = HRESULT_FROM_WIN32(GetLastError());
+            logf(log_verbosity::minimal, "Failed to get PATH environment variable [GetEnvironmentVariable]: %x\n", err);
```
MikhailArkhipov: It is, of course, unusual, but technically PATH variable doesn't have to exist...

src/main.cpp
```diff
@@ -183,6 +183,35 @@ namespace rhost {
...
+
+        std::wstring old_path_env(buff);
+        std::wstring new_path_env = r_dir.make_preferred().wstring() + L";" + old_path_env;
+        if (SetEnvironmentVariableW(L"PATH", new_path_env.c_str()) == 0) {
+            err = HRESULT_FROM_WIN32(GetLastError());
+            logf(log_verbosity::minimal, "Failed to set PATH environment variable [SetEnvironmentVariable]: %x\n", err);
+            std::terminate();
+        }
+#else
+        std::string old_path_env(std::getenv("LD_LIBRARY_PATH"));
```
MikhailArkhipov: If it doesn't exist, does it yield null or empty string?

src/main.cpp
```diff
@@ -183,6 +183,35 @@ namespace rhost {
...
+        std::string new_path_env = r_dir.make_preferred().string() + ":" + old_path_env;
+        if (setenv("LD_LIBRARY_PATH", new_path_env.c_str(), 1) == -1) {
```
MikhailArkhipov: Out of curiosity - `setenv` doesn't work in Win32?

karthiknadig: `setenv` is not supported in Win32.

src/main.cpp
karthiknadig: You are correct. I made an assumption that "PATH" is always available.

src/main.cpp
karthiknadig: I made an assumption that "PATH" is always available.

src/main.cpp
```diff
@@ -183,6 +183,40 @@ namespace rhost {
...
+#ifdef _WIN32
+        wchar_t buff[32767] = {}; // max buffer size for environment variable values on windows.
+        int nchars = GetEnvironmentVariable(L"PATH", buff, ARRAYSIZE(buff));
+        HRESULT err = S_OK;
+        if (nchars == 0) {
+            err = HRESULT_FROM_WIN32(GetLastError());
+            if (err == HRESULT_FROM_WIN32(ERROR_ENVVAR_NOT_FOUND)) {
+                buff[0] = '\0';
+            } else {
+                logf(log_verbosity::minimal, "Failed to get PATH environment variable [GetEnvironmentVariable]: %x\n", err);
```
int19h: `fatal_error` will log & terminate.

PR #142, opened 2017-06-09
MSBuild project for linux version of RHost

karthiknadig:


linux-build.proj
```diff
@@ -0,0 +1,48 @@
...
+
+  <Target Name="Restore">
+  </Target> 
+
+  <Target Name="Clean">
+    <Delete Files="$(OutputDirectory)$(RHostName)" />
+  </Target>
+
+  <Target Name="Rebuild">
+    <Exec Command="cd $(RHostDirectory)"/>
```
int19h: `Exec` has a `WorkingDirectory` attribute, so you can just set that directly.

linux-build.proj
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0"
+         DefaultTargets="Build"
+         xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <RHostName Condition=" '$(RHostName)'=='' ">Microsoft.R.Host</RHostName>
+    <RootDirectory Condition=" '$(RootDirectory)'=='' ">$(MSBuildThisFileDirectory)..\..\..\</RootDirectory>
```
int19h: It's weird that we have a project file here that makes assumption about the filesystem layout outside of what's in this repository. Effectively, this line makes this project file a build artifact for the larger RTVS build, rather than specifically for the host - if you check out just R-Host and try to build this, it'll fail. So it should probably go into the RTVS repo.

linux-build.proj
```diff
@@ -0,0 +1,48 @@
...
+    <OutputDirectory Condition=" '$(OutputDirectory)'=='' ">$(RootDirectory)bin\</OutputDirectory>
+    <IntermediateOutputDirectory Condition=" '$(IntermediateOutputDirectory)'=='' ">$(RootDirectory)obj\$(RHostName)\</IntermediateOutputDirectory>
```
int19h: In vcxproj, this currently also includes the platform:
```
    <IntDir>$(ObjDirectory)$(Configuration)\$(TargetName)\$(Platform)\Microsoft.R.Host</IntDir>
```
It's not really a big deal, but if we do build for multiple platforms, it means that intermediate object files won't be overwritten on every build if you switch back and forth.

linux-build.proj
```diff
@@ -0,0 +1,48 @@
...
+    <SourceDirectory Condition=" '$(SourceDirectory)'=='' ">$(MSBuildThisFileDirectory)src\</SourceDirectory>
+    
+
+    <Year Condition="'$(Year)' == ''">$([System.DateTime]::Now.Year)</Year>
+    <Date Condition="'$(Date)' == ''">$([System.DateTime]::Now.ToString("MMdd"))</Date>
+    <Time Condition="'$(Time)' == ''">$([System.DateTime]::Now.ToString("HHmm"))</Time>
+
+    <RHostDirectory Condition="'$(RHostDirectory)' == ''">$(MSBuildThisFileDirectory)</RHostDirectory>
+    <RHostSH Condition="'$(RHostSH)' == ''">$(RHostDirectory)build.sh</RHostSH>
+    <SolutionProperties Condition="'$(SolutionProperties)' == ''">
```
int19h: I thought this property is only needed if you're using `<Solution>` to build an .sln file - is there some other case for it?

linux-build.proj
int19h: Alternatively, if MSBuild on .NET Core supports solutions, we could use `SolutionDir`, like we do on Windows now.

linux-build.proj
```diff
@@ -0,0 +1,48 @@
...
+  </PropertyGroup>
+
+  <Target Name="Restore">
+  </Target> 
+
+  <Target Name="Clean">
+    <Delete Files="$(OutputDirectory)$(RHostName)" />
+  </Target>
+
+  <Target Name="Rebuild">
```
int19h: This is basically clean + rebuild, or it should be - so it can be implemented as `DependsOnTargets="Clean;Rebuild"` (and with empty body).

linux-build.proj
karthiknadig: SolutionDir works I will update this.

linux-build.proj
karthiknadig: This is not needed. Will remove.

PR #144, opened 2017-06-14
Fixes issue with help.

karthiknadig:


karthiknadig:
Since we had already discussed these changes I am checking this in.

PR #150, opened 2017-07-07
Capture R-host child processes' output and convert to message

karthiknadig:
https://github.com/Microsoft/RTVS/issues/3626

### DIFF
--- a/src/transport.cpp
+++ b/src/transport.cpp
@@ -28,11 +28,12 @@ using namespace rhost::protocol;
 namespace rhost {
     namespace transport {
         namespace {
-            const char devNull[] =
+            const int pipeSize = 4096;
 #ifdef _WIN32
-                "NUL";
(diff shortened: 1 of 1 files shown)


src/transport.cpp
```diff
@@ -85,6 +86,53 @@ namespace rhost {
...
             }
+
+            void read_stream_to_message(int fdr, const std::string& message_name) {
+                char line[pipeSize];
+                size_t len = pipeSize;
+                ssize_t nread;
+
+                for(;;) {
+#ifdef _WIN32
+                    nread = _read(fdr, line, pipeSize);
```
int19h: I wonder if you could just set up a bunch of global variables as aliases instead - like `const auto read = _read` etc.

int19h: On the other hand, since `pipe` call is different anyway, it's probably not worthwhile.

PR #156, opened 2017-10-02
OSX Build

MikhailArkhipov:
- adding zipconf.h as a backup since it is missing from homebrew libzip install


int19h:
Looks like that header is just installed in a non-standard location. If we use `pkg-config` to get build paths, we should pick it up.

https://github.com/Homebrew/legacy-homebrew/issues/13390

MikhailArkhipov:
OK, @karthiknadig can try on Ubuntu. I can check on macOS later.

int19h:
👍 if this also works on Linux (I don't see why it wouldn't, just need to check).

